PR #3, opened 2021-08-04
Device selection

jakub-dzon:
This PR introduces changes allowing users to select more than one device to run their workload.


api/v1alpha1/edgedeployment_types.go
```diff
@@ -41,18 +42,8 @@ const (
...
-type EdgeDeploymentPhase string
-
-const (
-	Deploying EdgeDeploymentPhase = "Deploying"
-	Running   EdgeDeploymentPhase = "Running"
-	Exited    EdgeDeploymentPhase = "Exited"
-)
-
 // EdgeDeploymentStatus defines the observed state of EdgeDeployment
 type EdgeDeploymentStatus struct {
```
pkliczewski, line 46: After this change deployment may run on one or more devices why not to use something similar to `deployment` status:

```yaml
status:
  availableReplicas: 2
  conditions:
  - lastTransitionTime: 2016-10-04T12:25:39Z
    lastUpdateTime: 2016-10-04T12:25:39Z
    message: Replica set "nginx-deployment-4262182780" is progressing.
    reason: ReplicaSetUpdated
    status: "True"
    type: Progressing
  - lastTransitionTime: 2016-10-04T12:25:42Z
    lastUpdateTime: 2016-10-04T12:25:42Z
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
  - lastTransitionTime: 2016-10-04T12:25:39Z
    lastUpdateTime: 2016-10-04T12:25:39Z
    message: 'Error creating: pods "nginx-deployment-4262182780-" is forbidden: exceeded quota:
      object-counts, requested: pods=1, used: pods=3, limited: pods=2'
    reason: FailedCreate
    status: "True"
    type: ReplicaFailure
  observedGeneration: 3
  replicas: 2
  unavailableReplicas: 2
```

jakub-dzon: IMO so complex status is not needed at this stage. 
I thought about keeping simple deployment status (as it was before that change) per device in the EdgeDeployment CR and decided to put it in the EdgeDevice because the reference from device to deployments was needed. Now I have second thoughts about it - labels on the EdgeDevice serve the purpose and maybe keeping per device  status in EdgeDeployment is a better option. WDYT about moving it back to the EdgeDeployment CR?

pkliczewski: sounds like a good idea

jakub-dzon: I remembered another reason for moving the deployment statuses to the EdgeDevice CR - potentially the list of devices in the EdgeDeployment CR could be massive in contrast to number of deployments per device, which should be rather small. 

I remember from the discussions that having long lists of devices for precise scheduling is not desired and here we would have additional data besides the IDs making it even bigger.

api/v1alpha1/edgedevice_types.go
```diff
@@ -64,10 +64,25 @@ type HardwareProfileConfiguration struct {
 
 // EdgeDeviceStatus defines the observed state of EdgeDevice
 type EdgeDeviceStatus struct {
-	Phase                     string      `json:"phase,omitempty"`
-	LastSeenTime              metav1.Time `json:"lastSeenTime,omitempty"`
-	LastSyncedResourceVersion string      `json:"lastSyncedResourceVersion,omitempty"`
-	Hardware                  *Hardware   `json:"hardware,omitempty"`
+	Phase                     string       `json:"phase,omitempty"`
```
masayag, line 67: What are the expected value of the Phase attribute ?

api/v1alpha1/edgedevice_types.go
```diff
@@ -64,10 +64,25 @@ type HardwareProfileConfiguration struct {
...
+	LastSeenTime              metav1.Time  `json:"lastSeenTime,omitempty"`
+	LastSyncedResourceVersion string       `json:"lastSyncedResourceVersion,omitempty"`
+	Hardware                  *Hardware    `json:"hardware,omitempty"`
+	Deployments               []Deployment `json:"deployments,omitempty"`
+}
+
+type EdgeDeploymentPhase string
+
+const (
+	Deploying EdgeDeploymentPhase = "Deploying"
```
masayag, line 77: none of those reflects an erroneous status, how should admin indicate if the EdgeDeployment failed to run ?

api/v1alpha1/edgedevice_types.go
```diff
@@ -64,10 +64,25 @@ type HardwareProfileConfiguration struct {
...
-	Phase                     string      `json:"phase,omitempty"`
-	LastSeenTime              metav1.Time `json:"lastSeenTime,omitempty"`
-	LastSyncedResourceVersion string      `json:"lastSyncedResourceVersion,omitempty"`
-	Hardware                  *Hardware   `json:"hardware,omitempty"`
+	Phase                     string       `json:"phase,omitempty"`
+	LastSeenTime              metav1.Time  `json:"lastSeenTime,omitempty"`
+	LastSyncedResourceVersion string       `json:"lastSyncedResourceVersion,omitempty"`
+	Hardware                  *Hardware    `json:"hardware,omitempty"`
+	Deployments               []Deployment `json:"deployments,omitempty"`
+}
```
masayag, line 72: shouldn't conditions be added as well, so the EdgeDevice can be watched ? (unless this is outside of the scope of this PR)

api/v1alpha1/edgedevice_types.go
jakub-dzon: The statuses are sent by the device agent and for now it is only `up`

api/v1alpha1/edgedevice_types.go
jakub-dzon: The data model accepts any string sent by the agent as a Phase and any erroneous status will be visible to the admin.

api/v1alpha1/edgedevice_types.go
jakub-dzon: I deliberately didn't implement any conditions to simplify the PoC. Deployment and Device state changes are reflected in `status.phase` and `status.deployments.phase`. 
IMO the conditions should be added at a later stage to provide detailed state information.

api/v1alpha1/edgedeployment_types.go
jakub-dzon: As per our discussion let's keep it as it is and adjust when user feedback instructs us to redesign.

controllers/edgedeployment_controller.go
```diff
@@ -64,71 +66,160 @@ func (r *EdgeDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
-	newEdgeDeployment := ownedDeployment.DeepCopy()
-	newEdgeDeployment.SetOwnerReferences(ownerRefs)
+	return nil
+}
 
-	return r.EdgeDeploymentRepository.Patch(ctx, ownedDeployment, newEdgeDeployment)
+func (r *EdgeDeploymentReconciler) addDeploymentsToDevices(ctx context.Context, name string, edgeDevices []managementv1alpha1.EdgeDevice) error {
+outer:
+	for _, edgeDevice := range edgeDevices {
```
masayag, line 171: running this one in parallel makes sense since we expect a lot of edge devices and minimize a failure of a single device to prevent updating the others.

controllers/edgedeployment_controller.go
```diff
@@ -64,71 +66,160 @@ func (r *EdgeDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		for _, deployment := range edgeDevice.Status.Deployments {
+			if deployment.Name == name {
+				// Move to the next device
+				continue outer
+			}
+		}
+		deploymentStatus := managementv1alpha1.Deployment{Name: name, Phase: managementv1alpha1.Deploying}
+		edgeDevice.Status.Deployments = append(edgeDevice.Status.Deployments, deploymentStatus)
+		err := r.EdgeDeviceRepository.UpdateStatus(ctx, &edgeDevice)
```
masayag, line 180: I think that `EdgeDeviceRepository.UpdateStatus` needs to fetch the device status first and then apply the change on it to minimize chances of failures (e.g. if 2 deployments were updated, or if the status is updated as a result of a heartbeat).

Else, we might face "cannot update resource, not latest version" kind of errors

controllers/edgedeployment_controller.go
```diff
@@ -64,71 +66,160 @@ func (r *EdgeDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			return nil, err
 		}
+		edgeDevices = append(edgeDevices, ed...)
 	}
-	return false
+	return edgeDevices, nil
+}
+
+func (r *EdgeDeploymentReconciler) getLabelledEdgeDevices(ctx context.Context, name, namespace string) ([]managementv1alpha1.EdgeDevice, error) {
+	selector := metav1.LabelSelector{MatchLabels: map[string]string{name: "true"}}
```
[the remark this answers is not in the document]
jakub-dzon: :+1: Will do.

controllers/edgedeployment_controller.go
```diff
@@ -64,71 +66,160 @@ func (r *EdgeDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 }
 
-// SetupWithManager sets up the controller with the Manager.
-func (r *EdgeDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&managementv1alpha1.EdgeDeployment{}).
-		Complete(r)
+func (r *EdgeDeploymentReconciler) finalizeRemoval(ctx context.Context, edgeDevices []managementv1alpha1.EdgeDevice, edgeDeployment *managementv1alpha1.EdgeDeployment) error {
+	for _, edgeDevice := range edgeDevices {
+		err := r.removeDeploymentFromDevice(ctx, edgeDeployment.Name, edgeDevice)
```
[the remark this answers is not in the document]
jakub-dzon: Parallelization is a great solution for potential performance problems with large number of devices to update. I'd like to postpone that optimization until we see such problems, though.

controllers/edgedeployment_controller.go
jakub-dzon: I'd like to postpone that optimization until we see performance issues.

controllers/edgedeployment_controller.go
```diff
@@ -64,71 +66,160 @@ func (r *EdgeDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				return ctrl.Result{Requeue: true}, err
+			}
+		}
+		return ctrl.Result{}, nil
 	}
 
-	if edgeDeployment.Status.Phase == "" {
-		edgeDeployment.Status.Phase = managementv1alpha1.Deploying
-		edgeDeployment.Status.LastTransitionTime = metav1.Now()
+	edgeDevices, err := r.getMatchingEdgeDevices(ctx, edgeDeployment)
```
pkliczewski, line 98: I think it would be good to skip if no changes to `Device` or `DeviceSelector` detected.

pkliczewski (commented):
What do you think about handling situation when `EdgeDevice` labels are updated which would cause workload being added/removed?

controllers/edgedeployment_controller.go
jakub-dzon: I don't think we can detect that without getting previous version of the CR. Besides, fetching the matching devices allows us to add/remove deployments to/from _devices_ that changed.

jakub-dzon:
> What do you think about handling situation when `EdgeDevice` labels are updated which would cause workload being added/removed?

The workload should be re-deployed when the deployment is reconciled and the label restored at the same time - the code actually won't work that way and needs to be fixed.

controllers/edgedeployment_controller.go
pkliczewski: I am not sure whether it would work in this case but there is a way to [filter](https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/) specific events based on new vs old CR.

controllers/edgedeployment_controller.go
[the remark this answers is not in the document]
jakub-dzon: I've created https://issues.redhat.com/browse/ECOPROJECT-221 to track the optimization-related changes

controllers/edgedeployment_controller.go
jakub-dzon: I've created https://issues.redhat.com/browse/ECOPROJECT-221 to track the optimization-related changes

controllers/edgedeployment_controller.go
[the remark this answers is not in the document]
jakub-dzon: Done

controllers/edgedeployment_controller.go
```diff
@@ -64,71 +66,160 @@ func (r *EdgeDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 	}
-	err = r.EdgeDeploymentRepository.UpdateStatus(ctx, edgeDeployment)
+
+	err = r.addDeploymentsToDevices(ctx, edgeDeployment.Name, edgeDevices)
 	if err != nil {
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	err = r.removeDeploymentFromNonMatchingDevices(ctx, edgeDeployment.Name, edgeDevices, labelledDevices)
+
```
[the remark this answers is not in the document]
jakub-dzon: Done.

controllers/edgedeployment_controller.go
```diff
@@ -64,71 +66,160 @@ func (r *EdgeDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+		return ctrl.Result{Requeue: true}, nil
 	}
 
-	err = r.addOwnerReference(ctx, edgeDevice, edgeDeployment)
+	labelledDevices, err := r.getLabelledEdgeDevices(ctx, edgeDeployment.Name, edgeDeployment.Namespace)
 	if err != nil {
-		return ctrl.Result{Requeue: true}, err
+		if !errors.IsNotFound(err) {
+			logger.Error(err, "Cannot retrieve Labelled Edge Deployments")
```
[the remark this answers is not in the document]
jakub-dzon: Done.

controllers/edgedeployment_controller.go
jakub-dzon: `EdgeDeviceRepository.UpdateStatus` can't just copy desired status to object freshly retrieved from etcd because the object coming in the parameter may have something in the status sub-resource that is gone in the latest repository version and we could restore it accidentally (especially `status.deployments` will be problematic). 
There are two options here: 
 - let the caller resolve conflict and then retry updating status
 - let's return error and make the reconciliation re-run

I went with the second option as it feel as a more natural way of dealing with problems in a controller.

controllers/edgedeployment_controller.go
jakub-dzon: I think filters should work in this case, but there is still the second argument for keeping it - a little better deployment consistency.

masayag (approved):

masayag:
> > What do you think about handling situation when `EdgeDevice` labels are updated which would cause workload being added/removed?
> 
> The workload should be re-deployed when the deployment is reconciled and the label restored at the same time - the code actually won't work that way and needs to be fixed.

This PR looks good to me, minus missing EdgeDevice labels management.
If you add support for EdgeDevice labels management to this PR, will it be possible adding it as a separate commit (just for easing the review) ?

jakub-dzon:
> > > What do you think about handling situation when `EdgeDevice` labels are updated which would cause workload being added/removed?
> > 
> > 
> > The workload should be re-deployed when the deployment is reconciled and the label restored at the same time - the code actually won't work that way and needs to be fixed.
> 
> This PR looks good to me, minus missing EdgeDevice labels management.
> If you add support for EdgeDevice labels management to this PR, will it be possible to add it as a separate commit (just for easing the review) ?

I will be working on it today and I will add it as a separate commit.

pkliczewski (approved):

jakub-dzon:
@masayag, @pkliczewski  - I've pushed two more commits addressing following issues:
- labels management;
- trying to update all of the EdgeDevices before returning error if one occurs;

Please review.

pkliczewski:
Looks good to me. @masayag feel free to merge when happy

masayag (approved):

PR #4, opened 2021-08-05
Allow changing deployment configuration from configmap

masayag:
This PR introduce reading configuration from a configmap which is part
of the operator resources.

In order to change container's http-server port, the user can update the
value of the HTTP_PORT field, followed by a restart of the deployment,
e.g.:

```
kubectl patch cm -n k4e-operator-system k4e-operator-manager-config \
        --type merge --patch '{"data":{"HTTP_PORT": "7776"}}'

kubectl rollout restart deployment -n k4e-operator-system k4e-operator-controller-manager
```

In addition, the service resource will have to be changed accordingly:

```
kubectl patch svc -n k4e-operator-system k4e-operator-controller-manager \
        --type merge -p '{"spec": {"ports": [{"port": 7776,"targetPort": 8888,"name": "http"}]}}'
```

---
In order to change the port of the service, only the targetPort of the
service resource needs to be updated:

```
kubectl patch svc -n k4e-operator-system k4e-operator-controller-manager \
        --type merge -p '{"spec": {"ports": [{"port": 7776,"targetPort": 8888,"name": "http"}]}}'
kubectl port-forward service/k4e-operator-controller-manager -n k4e-operator-system 7776
```

The `ygdd` agent needs to be started with
`--http-server=localhost:7776`:
```
sudo go run ./cmd/yggd --log-level trace --transport http --cert-file /etc/pki/consumer/cert.pem \
     --key-file /etc/pki/consumer/key.pem --client-id-source machine-id \
     --http-server 127.0.0.1:7776
```


main.go
```diff
@@ -141,7 +141,22 @@ func main() {
 		if err != nil {
 			setupLog.Error(err, "cannot start http server")
 		}
-		address := fmt.Sprintf(":%v", Port)
+		var cm v1.ConfigMap
+		cmKey := client.ObjectKey{Namespace: "k4e-operator-system", Name: "k4e-operator-manager-config"}
```
jakub-dzon, line 145: What do you think about using the config map as env variables source in the deployment, like here: https://github.com/openshift/assisted-service/blob/04326d154f72cb71ece287dad19b7f016464ae88/deploy/assisted-service.yaml#L40?

pkliczewski: I wonder whether it would be possible to parametrize namespace.

masayag: with Jakub's suggestion it won't be needed

pkliczewski (approved):
/lgtm

jakub-dzon (approved):

PR #6, opened 2021-08-31
Add storage configuration for edge device

masayag:
The edge-device configuration is extended with stoarge
configuration, to allow edge devices to connect to via
S3 to the storage and upload (or even download ?) from it.


api/v1alpha1/edgedevice_types.go
```diff
@@ -62,6 +65,14 @@ type HardwareProfileConfiguration struct {
 	Scope string `json:"scope,omitempty"`
 }
 
+type StorageConfiguration struct {
```
jakub-dzon, line 68: WDYT about nesting the s3-specific configuration to allow for adding other types in the future?
Like:
```yaml
storage:
  s3:
    bucket_host:  ...
    ...
  other_storage_type:
    ...
```

api/v1alpha1/edgedevice_types.go
```diff
@@ -62,6 +65,14 @@ type HardwareProfileConfiguration struct {
...
+	BucketHost         string `json:"bucket_host,omitempty"`
```
jakub-dzon, line 69: Please use camelCase in json property names for consistency.

swagger.yaml
```diff
@@ -172,6 +174,20 @@ definitions:
           - full
           - delta
 
+  storage-configuration:
```
jakub-dzon, line 177: Please nest s3-specific configuration to allow for easy extension with other storage types.

jakub-dzon (changes requested):

api/v1alpha1/edgedevice_types.go
masayag: there is no consistency in the file :-)

api/v1alpha1/edgedevice_types.go
jakub-dzon: It might be sftp server, or network drive for example.

jakub-dzon (approved):

pkliczewski (approved):

PR #9, opened 2021-09-07
Deployment data transfer config

jakub-dzon:
This PR introduces data model for specifying data paths to synchronize between edge device and control plane storage and logic for storing last data upload time.


swagger.yaml
```diff
@@ -417,6 +439,9 @@ definitions:
     properties:
       name:
         type: string
+      last_data_upload:
+        type: string
+        format: date-time
```
pkliczewski, line 444: What is the value of tracking when data sync occurred?

jakub-dzon: So the admin/user would know whether data they have in the bucket are up to date. They may see old timestamps in the storage and files still can be up to date and the time in CR will confirm that.

pkliczewski: Once they have that knowledge what mechanisms do we plan for them to react if something is not right?

jakub-dzon: If they have access to the device, they can check the device logs, see what's the problem is and fix it; most likely it would be network or remote storage problems.

internal/yggdrasil/yggdrasil.go
```diff
@@ -245,6 +245,7 @@ func (h *Handler) updateDeploymentStatuses(oldDeployments []v1alpha1.Deployment,
 				deployment.Phase = v1alpha1.EdgeDeploymentPhase(status.Status)
 				deployment.LastTransitionTime = metav1.Now()
 			}
+			deployment.LastDataUpload = metav1.NewTime(time.Time(status.LastDataUpload))
```
masayag, line 248: Shouldn't the assignment of value for `LastDataUpload` depends on the existence of data-paths for the edge deployment ?

jakub-dzon: There is no need. When the data transfer doesn't take place the `status.LastDataUpload` is set to `time.Time` zero value, which in turn doesn't get stored in the CR.

masayag (approved):

pkliczewski (approved):

PR #15, opened 2021-09-29
change JSON field names to camelCase

ydayagi:
https://issues.redhat.com/browse/ECOPROJECT-253

### DIFF
--- a/docs/http-api.md
+++ b/docs/http-api.md
@@ -1,3 +1,34 @@
 # HTTP API
 
-TODO
+Operator's HTTP API consists of four endpoints used in communication with yggdrasil/agent:
+
+ - `GET /data/{device_id}/in`
+ - `POST /data/{device_id}/out`
+ - `GET /control/{device_id}/in`
+ - `POST /control/{device_id}/out`
+
+## `GET /data/{device_id}/in`
+
+This endpoint is used by the agent to retrieve its expected configuration; the response is `message` object described in the [Swagger specification](http_api_swagger.md). 
+The `content` field of the message contains payload understood by the device-worker. The payload is described by the `device-configuration-message` Swagger object. 
+
+The `content` is forwarded to the `device-worker` and processed there.
+
+
+## `POST /data/{device_id}/out` 
+
+This endpoint is used by the agent to send information to the operator. Currently, there are only two types of message contents supported by this endpoint (see [Swagger specification](http_api_swagger.md)):
+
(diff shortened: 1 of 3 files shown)


jakub-dzon (changes requested):
Please run `make generate manifests` to update generated files.

Make sure that https://github.com/jakub-dzon/k4e-operator/blob/35f1efee0d5f61e376db6e5d81e7028443a128ce/internal/yggdrasil/yggdrasil.go#L259  
works correctly after the changes; I think it might be broken.

swagger.yaml
```diff
@@ -26,14 +26,14 @@ securityDefinitions:
...
-  /data/{device_id}/in:
+  /data/{deviceId}/in:
     get:
       operationId: GetDataMessageForDevice
       tags:
         - yggdrasil
       parameters:
         - in: path
-          name: device_id
+          name: deviceId
```
jakub-dzon, line 36: Changes in swagger break compatibility with device-worker, which sends payload in that model.

ydayagi: Are u saying I should not change swagger?

ydayagi:
> Please run `make generate manifests` to update generated files.
> 
> Make sure that
> 
> https://github.com/jakub-dzon/k4e-operator/blob/35f1efee0d5f61e376db6e5d81e7028443a128ce/internal/yggdrasil/yggdrasil.go#L259
> 
> 
> works correctly after the changes; I think it might be broken.

u mean device-worker and not Yggdrasil, correct?

swagger.yaml
jakub-dzon: Yes. Unless you can provide compatibility with the older model.

jakub-dzon:
> > Please run `make generate manifests` to update generated files.
> > Make sure that
> > https://github.com/jakub-dzon/k4e-operator/blob/35f1efee0d5f61e376db6e5d81e7028443a128ce/internal/yggdrasil/yggdrasil.go#L259
> > 
> > works correctly after the changes; I think it might be broken.
> 
> u mean device-worker and not Yggdrasil, correct?

Operator. 
device-worker sends hardware information to operator in model defined by swagger.yaml and https://github.com/jakub-dzon/k4e-operator/blob/35f1efee0d5f61e376db6e5d81e7028443a128ce/internal/yggdrasil/yggdrasil.go#L259 function maps swagger model to CR model. All ` utils.Copy` calls might be problematic, because they marshal one data type to JSON and unmarshall it to another.

PR #19, opened 2021-09-30
Retry updating the EdgeDevice status

masayag:
The PR uses the patch method for updating the edge-device status.
In addition, it retries 3 times in case of a failure when the device
posts its information to the operator.

ECOPROJECT-255

### DIFF
--- a/controllers/edgedeployment_controller.go
+++ b/controllers/edgedeployment_controller.go
@@ -140,8 +140,9 @@ func (r *EdgeDeploymentReconciler) removeDeploymentFromDevice(ctx context.Contex
 			newDeployments = append(newDeployments, deployment)
 		}
 	}
+	patch := client.MergeFrom(edgeDevice.DeepCopy())
 	edgeDevice.Status.Deployments = newDeployments
-	err := r.EdgeDeviceRepository.UpdateStatus(ctx, &edgeDevice)
+	err := r.EdgeDeviceRepository.PatchStatus(ctx, &edgeDevice, &patch)
 	if err != nil {
 		return err
 	}
@@ -184,8 +185,9 @@ func (r *EdgeDeploymentReconciler) addDeploymentsToDevices(ctx context.Context, 
 	for _, edgeDevice := range edgeDevices {
 		if !hasDeployment(edgeDevice, name) {
 			deploymentStatus := managementv1alpha1.Deployment{Name: name, Phase: managementv1alpha1.Deploying}
+			patch := client.MergeFrom(edgeDevice.DeepCopy())
 			edgeDevice.Status.Deployments = append(edgeDevice.Status.Deployments, deploymentStatus)
-			err := r.EdgeDeviceRepository.UpdateStatus(ctx, &edgeDevice)
+			err := r.EdgeDeviceRepository.PatchStatus(ctx, &edgeDevice, &patch)
 			if err != nil {
 				errs = append(errs, err)
 				continue
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -18,6 +18,7 @@ import (
 	operations "github.com/jakub-dzon/k4e-operator/restapi/operations/yggdrasil"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"time"
 )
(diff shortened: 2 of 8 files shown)


internal/yggdrasil/yggdrasil.go
```diff
@@ -234,6 +239,29 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
...
+func (h *Handler) updateDeviceStatus(ctx context.Context, device *v1alpha1.EdgeDevice, updateFunc func(d *v1alpha1.EdgeDevice)) middleware.Responder {
+	patch := client.MergeFrom(device.DeepCopy())
+	updateFunc(device)
+	err := h.deviceRepository.PatchStatus(ctx, device, &patch)
+	if err == nil {
+		return nil
+	}
+
+	// retry patching the edge device status
+	for i := 0; i < 3; i++ {
```
jakub-dzon, line 251: What do you think about pausing between retries? 
Exponential back-off might be helpful if there is a lot of changes happening to the device at the same time.

internal/yggdrasil/yggdrasil.go
```diff
@@ -168,7 +169,10 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 	case "heartbeat":
 		heartbeat := models.Heartbeat{}
 		contentJson, _ := json.Marshal(msg.Content)
-		json.Unmarshal(contentJson, &heartbeat)
+		err := json.Unmarshal(contentJson, &heartbeat)
+		if err != nil {
+			return nil
```
jakub-dzon, line 174: That should be an error (HTTP 400)  instead of `nil`

internal/yggdrasil/yggdrasil.go
```diff
@@ -195,7 +200,10 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 			if errors.IsNotFound(err) {
 				contentJson, _ := json.Marshal(msg.Content)
 				registrationInfo := models.RegistrationInfo{}
-				json.Unmarshal(contentJson, &registrationInfo)
+				err := json.Unmarshal(contentJson, &registrationInfo)
+				if err != nil {
+					return nil
```
jakub-dzon, line 205: That also should be HTTP 400 instead of `nil`

internal/yggdrasil/yggdrasil.go
```diff
@@ -234,6 +239,29 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
...
+	updateFunc(device)
+	err := h.deviceRepository.PatchStatus(ctx, device, &patch)
+	if err == nil {
+		return nil
+	}
+
+	// retry patching the edge device status
+	for i := 0; i < 3; i++ {
+		device2, err := h.deviceRepository.Read(ctx, device.Name, device.Namespace)
+		if err == nil {
```
jakub-dzon, line 253: What do you think about inverting the condition to eliminate nesting?

controllers/edgedeployment_controller.go
```diff
@@ -184,8 +185,9 @@ func (r *EdgeDeploymentReconciler) addDeploymentsToDevices(ctx context.Context,
 	for _, edgeDevice := range edgeDevices {
 		if !hasDeployment(edgeDevice, name) {
 			deploymentStatus := managementv1alpha1.Deployment{Name: name, Phase: managementv1alpha1.Deploying}
+			patch := client.MergeFrom(edgeDevice.DeepCopy())
 			edgeDevice.Status.Deployments = append(edgeDevice.Status.Deployments, deploymentStatus)
-			err := r.EdgeDeviceRepository.UpdateStatus(ctx, &edgeDevice)
```
jakub-dzon, line 188: `UpdateStatus` can be removed from the `EdgeDeviceRepository`; let's not leave too many options for updating status.

jakub-dzon (changes requested):

internal/yggdrasil/yggdrasil.go
masayag: indeed. for sure not nil. I added 400 to swagger.

pkliczewski (commented):
/lgtm

pkliczewski (approved):

internal/yggdrasil/yggdrasil.go
```diff
@@ -177,63 +181,92 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
...
-				device.Namespace = h.initialNamespace
-				device.Spec.OsImageId = registrationInfo.OsImageID
-				device.Finalizers = []string{YggdrasilConnectionFinalizer, YggdrasilWorkloadFinalizer}
-				err := h.deviceRepository.Create(ctx, &device)
-				if err != nil {
-					logger.Error(err, "Cannot save EdgeDevice")
-					return operations.NewPostDataMessageForDeviceInternalServerError()
-				}
-
+			if !errors.IsNotFound(err) {
```
jakub-dzon, line 200: :+1:

jakub-dzon (approved):

PR #25, opened 2021-10-20
Testing: Added edgedeployment controller

eloycoto:
Hi, 

This PR is a bit big, sorry, I tried to keep as small as possible, but to many things to add:

- Added gomock and  mockery because it's needed by swagger. 
- Added interfaces for internal/repository so I can test the controllers. 
- Refactor the ginkgo test-suite and edgedevices_test, so things are easy to test in this way. 
- Added 15 new test on edgedeployment controller test, a few tests are missing, but want to get feedback. 

 Regards


controllers/edgedeployment_controller_test.go
```diff
@@ -0,0 +1,509 @@
...
+
+				// when
+				res, err := edgeDeploymentReconciler.Reconcile(context.TODO(), req)
+
+				// then
+				Expect(err).NotTo(HaveOccurred())
+				Expect(res).To(Equal(reconcile.Result{Requeue: true, RequeueAfter: 0}))
+			})
+
+			It("Added finalizer failed requeue with error", func() {
```
pkliczewski, line 149: What do you think about adding delete device and see how finalizers are removed? Can be done in separate PR

internal/repository/edgedeployment/edgedeployment.go
```diff
@@ -2,20 +2,31 @@ package edgedeployment
 
 import (
 	"context"
+
 	"github.com/jakub-dzon/k4e-operator/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/fields"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	_ "github.com/golang/mock/mockgen/model"
```
masayag, line 10: is it needed here ? will the compilation or code generation fails without it?

controllers/edgedeployment_controller_test.go
```diff
@@ -0,0 +1,509 @@
...
+				// when
+				res, err := edgeDeploymentReconciler.Reconcile(context.TODO(), req)
+
+				// then
+				Expect(err).To(HaveOccurred())
+				Expect(res).To(Equal(reconcile.Result{Requeue: true, RequeueAfter: 0}))
+			})
+		})
+
+		Context("devices section", func() {
```
masayag, line 166: Is this context mainly about using device selector ?
perhaps another context should be for using the device so we'll get full cover of `getMatchingEdgeDevices()`

PR #33, opened 2021-10-29
CI: Add coverage check

eloycoto:
On a new PR, a new check will be added with the percent of the test
coverage, it'll fail if the coverage is smaller than 80%.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,5 +1,6 @@
 name: Go
-
+permissions:
+  actions: write
 on:
   push:
     branches: [ main ]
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -22,9 +22,19 @@ jobs:
...
+        make test-create-coverage
+        COVERAGE=$(go tool cover --func=cover.out |  grep total | grep -Eo '[0-9]+\.[0-9]+')
+        echo "testCoverage=${COVERAGE}" >> $GITHUB_ENV
 
     - uses: actions/upload-artifact@v2
       with:
         name: test-coverage
         path: coverage.html
+
+    - uses: ouzi-dev/commit-status-updater@v1.1.0
```
pkliczewski, line 35: build fails with:
` Error: error while setting context status: Resource not accessible by integration`

eloycoto: yes, yes, I was aware of that, I need to add a credential ;-)

PR #34, opened 2021-11-06
Enable setting of log verbosity level 

gciavarrini:


main.go
```diff
@@ -84,6 +86,14 @@ func init() {
...
 	utilruntime.Must(routev1.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
+
+	opts := zap.Options{}
+	opts.BindFlags(flag.CommandLine)
+	flag.Parse()
+	logger := zap.New(zap.UseFlagOptions(&opts))
+	ctrl.SetLogger(logger)
+
+	setupLog = ctrl.Log.WithName("setup")
```
masayag, line 96: Is there a need to preserve the same call on line 59 in this file ?

main.go
```diff
@@ -84,6 +86,14 @@ func init() {
 	utilruntime.Must(obv1.AddToScheme(scheme))
 	utilruntime.Must(routev1.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
+
+	opts := zap.Options{}
+	opts.BindFlags(flag.CommandLine)
```
masayag, line 91: when invoked from a container, we don't pass any command-line arguments.
in k8s, we should allow the user to set verbosity via configmap of the operator.
hence by changing the value in the config map and restarting the operator, we should allow the user to change verbosity level.
IMO we should prefer env-vars vs cli args, so we'll have the same behavior for the app when it runs locally via `make run` or when it runs in a pod.
Have a look at `OBC_AUTO_CREATE`

masayag (changes requested):

main.go
gciavarrini: Yes, it makes sense. I'll change it. Thanks

gciavarrini:
@masayag: I modified the PR according to your suggestion. I've also added a watcher.
I'm not sure about the `getConfigMapName` function.

main.go
```diff
@@ -164,10 +203,117 @@ func main() {
...
+	}
+
+	currentConfigMap, err := k8sClient.CoreV1().ConfigMaps(operatorNamespace).Get(context.TODO(), configMapName, metav1.GetOptions{})
+	if err != nil {
+		setupLog.Error(err, "Unable to get ConfigMap", "namespace", operatorNamespace)
+		os.Exit(1)
+	}
+	setupLog.V(1).Info("operator configmap found", "operatorNamespace", operatorNamespace, "configmap name", configMapName)
+
+	mutex = &sync.Mutex{}
```
masayag, line 238: why is this mutex needed ? Seems it can be defined from `WatchForChanges` and doesn't have to be passed to it.

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,78 @@
...
+
+func checkConfigMapChanges(eventChannel <-chan watch.Event, isLocalRun bool, dataField string, dataValue string, mutex *sync.Mutex) {
+	for {
+		event, open := <-eventChannel
+		if open {
+			switch event.Type {
+			case watch.Added:
+				fallthrough
+			case watch.Modified:
+				mutex.Lock()
```
masayag, line 39: the pattern for using locks requires `defer mutex.Unlock()` right after the lock is acquired, else, the exit statement on line 59 will break without releasing the lock.

main.go
```diff
@@ -164,10 +203,117 @@ func main() {
...
+	go watchers.WatchForChanges(k8sClient, Config.LocalRun, operatorNamespace, configMapName, "VERBOSITY", currentConfigMap.Data["VERBOSITY"], mutex)
```
masayag, line 239: `"VERBOSITY"` is repeated several times, perhaps it can be defined as constant ?

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,78 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+func WatchForChanges(clientset *kubernetes.Clientset, isLocalRun bool, namespace string, mgrConfigMapName string, dataField string, dataValue string, mutex *sync.Mutex) {
+	logger := log.FromContext(context.TODO())
+	logger.V(0).Info("watchForChanges", "namespace", namespace, "data field", dataField, "current value", dataValue)
+	logger.V(1).Info("watchForChanges", "namespace", namespace, "data field", dataField, "current value", dataValue)
```
masayag, line 19: one of the two lines (line 18 or 19) isn't needed

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,78 @@
...
+	for {
+		watcher, err := clientset.CoreV1().ConfigMaps(namespace).Watch(context.TODO(),
```
masayag, line 21: Is `Watch` a blocking call ?

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,78 @@
...
+				fallthrough
+			case watch.Modified:
+				mutex.Lock()
+				if updatedMap, ok := event.Object.(*corev1.ConfigMap); ok {
+					if updatedValue, ok := updatedMap.Data[dataField]; ok {
+						if updatedValue != dataValue {
+							if isLocalRun {
+								return
+							}
+							//restart pod
```
masayag, line 46: Can I assume that the watcher is an alternative for restarting the pod?

masayag (changes requested):

main.go
```diff
@@ -18,44 +18,59 @@ package main
...
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	//+kubebuilder:scaffold:imports
 )
 
 const (
-	initialDeviceNamespace = "default"
+	initialDeviceNamespace   = "default"
+	defaultOperatorNamespace = "k4e-operator-system"
+	mgrConfigMapName         = "k4e-operator-manager-config" //TODO: fix me
```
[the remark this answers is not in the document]
gciavarrini: I was not sure about the default value. If it's ok, I can remove the comment.

README.md
```diff
@@ -25,3 +25,13 @@ Run `make run` to start the operator.
 The port used by the operator HTTP API (`8888`) has to be available outside the cluster, so following port-forwarding command needs to be executed before attempting to communicate with it:
 
 `kubectl port-forward service/k4e-operator-controller-manager -n k4e-operator-system 8888`
+
+## Change log verbosity
+
+In order to change the log verbosity, the user can update the value of the `VERBOSITY` field.
```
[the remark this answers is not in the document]
gciavarrini: Yes, good suggestion.

main.go
```diff
@@ -164,10 +203,117 @@ func main() {
...
+	operatorNamespace, err := getOperatorNamespace(Config.LocalRun)
+	if err != nil {
+		setupLog.Error(err, "Cannot get the operator namespace")
+		os.Exit(1)
+	}
+	setupLog.V(1).Info("operator namespace found", "operatorNamespace", operatorNamespace)
+
+	configMapName, err := getConfigMapName(k8sClient, operatorNamespace, Config.LocalRun)
+	if err != nil {
+		setupLog.Error(err, "operator namespace", operatorNamespace)
```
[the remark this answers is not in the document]
gciavarrini: Is there any case in which the name of the ConfigMap is changed by the user? I created this function to cover this case, but I'm not sure if it's useful or not.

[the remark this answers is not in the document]
masayag: I don't think so. It is part of the operator settings. 
Also, k8s resource names are immutables. Removing the configmap might cause some issues.

main.go
masayag: Another thought about `VERBOSITY`, AFAIK `LOG_LEVEL` is commonly used as a config-value name to represent the verbosity of the logger.
Would you mind changing it ? `VERBOSITY` by itself might not be specific enough to refer only to logs.

main.go
```diff
@@ -164,10 +203,117 @@ func main() {
...
+
 	setupLog.Info("starting manager")
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
+}
+
+func getK8sClientSet() (*kubernetes.Clientset, error) {
+	if Config.LocalRun {
```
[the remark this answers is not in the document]
gciavarrini: The manual stop&restart is not related to the LocalRun option. 
The LocalRun option discriminates between in cluster and out of the cluster run because the kubernetes ClientSet must be retrieved differently int the two cases (otherwise all the ClientSet will be `nil`). Is there any way to implicitly get the in/out cluster run? Or am I misunderstanding something?

watchers/configmap_watcher.go
gciavarrini: IIUC, `Watch` is a blocking call. It waits for events on the `Event` channel. 
Is there anything you're not convinced of?

watchers/configmap_watcher.go
gciavarrini: During the rebase I forgot to delete the comment: sorry about that.
I tried to refresh at runtime the log level but it doesn't work. At the moment, restarting the pod seems the most feasible way to go.

main.go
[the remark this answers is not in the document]
masayag: How about do nothing if `Config.LocalRun` is true ?
In dev-mode the value of the verbosity can be provided by the CLI, unlike in a production env.
We can agree that for log-level change, the developer will have to stop & start after setting new value.
Creating the watcher will be done only for production (non local run).

Then we won't have to support dev-mode code in production env.

watchers/configmap_watcher.go
masayag: this call is running in an infinite loop (line 20). When the watcher is initialized, it returns the `watcher.ResultChan()`.
Next, there is another infinite for loop (line 32) that reads from that channel. 

My question is why the outer infinite loop is required ? In which case would we need to re-initialize the watcher ?

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,78 @@
...
+		if open {
+			switch event.Type {
+			case watch.Added:
+				fallthrough
+			case watch.Modified:
+				mutex.Lock()
+				if updatedMap, ok := event.Object.(*corev1.ConfigMap); ok {
+					if updatedValue, ok := updatedMap.Data[dataField]; ok {
+						if updatedValue != dataValue {
+							if isLocalRun {
```
jakub-dzon, line 43: I think we shouldn't even start the watcher if we're running in "local mode".

config/rbac/leader_election_role.yaml
```diff
@@ -4,6 +4,13 @@ kind: Role
 metadata:
   name: leader-election-role
 rules:
+- apiGroups:
+  - ""
+  resources:
```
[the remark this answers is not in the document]
gciavarrini: This is needed for [getConfigMapName](https://github.com/gciavarrini/k4e-operator/blob/e50b2c7214cf978c09a111a7f2cddf30b1f603d1/main.go#L299). I tried to add this rule in `config/rbac/role.yaml` but it didn't work.

However, since Moti suggested eliminating this function and using a constant, this change is no longer necessary.

watchers/configmap_watcher.go
gciavarrini: The `Watch` function is implemented by a `Get` call to the Kubernetes API server. So the `Watch` is a blocking call but the API server may close the close due to a timeout. So the infinite loop in line 20 handles this case. Is that reasonable for you?

masayag: yes. thanks for the explanation.

main.go
```diff
@@ -164,10 +203,117 @@ func main() {
 		setupLog.Info("starting http server", "address", address)
 		log.Fatal(http.ListenAndServe(address, h))
 	}()
+
+	var mutex *sync.Mutex
+
+	k8sClient, err := getK8sClientSet()
```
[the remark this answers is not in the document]
gciavarrini: I'm using the `mgr.GetConfig()` as you suggest. The `Get()` method from client `mgr.GetClient()` requires a `key ObjectKey` to get ConfigMaps but I don't' how to get it.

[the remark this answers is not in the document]
gciavarrini: Thanks! I'll fix it.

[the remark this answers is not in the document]
gciavarrini: There may be another problem. It seems to me that `client.Client` from `mgr.getClient()` doesn't have a `Watch` function. What do you think?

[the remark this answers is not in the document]
jakub-dzon: I see that you've already used `k8sClient, err := kubernetes.NewForConfig(mgr.GetConfig())` to create your client and limited watcher to be started only in in-cluster scenario; I don't think this discussion is relevant anymore:)

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,60 @@
...
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var logger = log.FromContext(context.TODO())
```
masayag, line 15: WDYT of replacing this logger with adding passing the logger (setupLog) from the main to this function ?

gciavarrini: The `setupLog` has name "setup"  ([main@108](https://github.com/jakub-dzon/k4e-operator/blob/338ec3af0d0ecc40626d90fae00c65b3d77bf487/main.go#L108)).
Is this ok for logger in `configmap_watcher.go`?

masayag: one might claim that setting a watcher for the configmap is part of the operator's setup.

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,60 @@
...
+func checkConfigMapChanges(eventChannel <-chan watch.Event, dataField string, dataValue string, mutex *sync.Mutex) {
+	for {
+		event, open := <-eventChannel
+		if open {
+			switch event.Type {
+			case watch.Added:
+				fallthrough
+			case watch.Modified:
+				mutex.Lock()
+				defer mutex.Unlock()
```
jakub-dzon, line 40: defer function is executed at the end of surrounding function; in this case it will (almost) never be executed, as the function returns only whe the eventChannel is closed and the infinite loop is broken.
One solution is to wrap whatever the whole case in an anonymous function.

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,60 @@
...
+				if updatedMap, ok := event.Object.(*corev1.ConfigMap); ok {
+					if updatedValue, ok := updatedMap.Data[dataField]; ok {
+						if updatedValue != dataValue {
+							logger.Info("restarting pod to update the logging level", "current level", dataValue, "new level", updatedValue)
+							os.Exit(1)
+						}
+					}
+				}
+				mutex.Unlock()
```
jakub-dzon, line 49: Either that or properly handled defer at line 40; not both.

jakub-dzon (changes requested):

watchers/configmap_watcher.go
masayag: rethinking the need for this mutex: the main function is executed in a single thread.
i don't think there will be a concurrent access here.
wdyt of removing the mutex entirely ?

jakub-dzon: I agree - we don't access or modify any shared state; the only reaction is `os.Exit(1)` if any.

jakub-dzon (approved):

pkliczewski (approved):

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func WatchForChanges(clientset *kubernetes.Clientset, namespace string, configMapName string, dataField string, dataValue string, logger logr.Logger) {
+	logger.V(1).Info("watch for changes", "namespace", namespace, "configMap name", configMapName, "data field", dataField, "current value", dataValue)
+	for {
+		watcher, err := clientset.CoreV1().ConfigMaps(namespace).Watch(context.TODO(),
+			metav1.SingleObject(metav1.ObjectMeta{Name: configMapName, Namespace: namespace}))
+		if err != nil {
+			logger.Error(err, "cannot create watcher", "namespace", namespace, "configMap name", configMapName)
+			os.Exit(1)
```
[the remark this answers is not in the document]
jakub-dzon: As for `Exit` in the error handling logic, I think retrying to create the watcher with some delay should be sufficient (`sleep`+`continue`); we don't need to crash the pod. 

Exit in line 39 is the actual logic that should remain as it is to restart the operator when configuration changes.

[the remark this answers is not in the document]
masayag: Agreed. @gciavarrini can you update the pr accordingly ?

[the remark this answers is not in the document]
gciavarrini: Yes, but can a fixed delay be sufficient? Do you think it's necessary to think about max retry attempts or backoff delay?
I'm not familiar with this network context, so I don't know how to calibrate the delay (any suggestion?). Should it be a parameter?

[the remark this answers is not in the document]
jakub-dzon: I'd go with exponential back-off.

watchers/configmap_watcher.go
```diff
@@ -0,0 +1,88 @@
...
+			}
+		} else {
+			// If eventChannel is closed, it means the server has closed the connection
+			return
+		}
+	}
+}
+
+func createWatcher(clientset kubernetes.Interface, namespace string, configMapName string, watcher *watch.Interface) error {
+	err := retry.OnError(backoff, retriable, func() error {
```
jakub-dzon, line 72: Good finding! I didn't know there is already an utility for that in the apimachinery.

jakub-dzon (approved):

masayag (approved):

PR #38, opened 2021-11-16
Run edge devices updates concurrently

ydayagi:
https://issues.redhat.com/browse/ECOPROJECT-232


controllers/edgedeployment_controller.go
```diff
@@ -243,6 +293,29 @@ func (r *EdgeDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+func (r *EdgeDeploymentReconciler) ExecuteConcurrent(funcs []concurrentFunc) []error {
```
jakub-dzon, line 296: There is no need for this to be public here. You can create separate package responsible for concurrent execution and there it would make sense to have it public. It might even be useful for other packages.

controllers/edgedeployment_controller.go
```diff
@@ -285,3 +358,25 @@ func mergeErrorMessages(errs []error) string {
 	}
 	return message
 }
+
+func SplitEdgeDevices(s []managementv1alpha1.EdgeDevice, x int) [][]managementv1alpha1.EdgeDevice {
```
jakub-dzon, line 362: There is no need for this to be public (here). Maybe moving it to utils or specialized device utils would be an option? 
Also, there is nothing in the function that actually needs `EdgeDevice` so maybe  using `interface{}` in parameter and return value would be possible?

controllers/edgedeployment_controller.go
```diff
@@ -285,3 +358,25 @@ func mergeErrorMessages(errs []error) string {
...
+	if x <= 0 {
+		return nil
```
jakub-dzon, line 364: That should be an error.

controllers/edgedeployment_controller_test.go
```diff
@@ -630,5 +631,24 @@ var _ = Describe("Controllers", func() {
 				Expect(res).To(Equal(reconcile.Result{Requeue: true, RequeueAfter: 0}))
 			})
 		})
+
+		Context("Concurrency", func() {
+			It("SplitEdgeDevices", func() {
+				const (
+					nEdgeDevices = 100
```
jakub-dzon, line 638: More input cases needed.
I would cover:
- less devices than concurrency;
- exactly the same number of devices as concurrency;
- exactly multiple of the concurrency number of devices (nConcurrency=7; nEdgeDevices=14);
- negative concurrency;
- concurrency zero;
- concurrency one;
- number of devices equals multiple of concurrency-1 (nConcurrency=7; nEdgeDevices=13);

controllers/edgedeployment_controller.go
```diff
@@ -243,6 +293,29 @@ func (r *EdgeDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
+func (r *EdgeDeploymentReconciler) ExecuteConcurrent(funcs []concurrentFunc) []error {
+	nFuncs := len(funcs)
+	if nFuncs == 0 {
+		return nil
+	}
+	returnValues := make([][]error, nFuncs)
+	var wg sync.WaitGroup
+	wg.Add(nFuncs)
+	for i := 0; i < nFuncs; i++ {
```
pkliczewski, line 304: What do you think about about not using slice with functions? We iterate over inputs to create it and later we iterate again to run it. We could only have one loop when we run it.

controllers/edgedeployment_controller.go
ydayagi: it was just temporary. I will add it to Utils.

controllers/edgedeployment_controller.go
ydayagi: It's only in theory. It is an internal function to be used in this file. I just did that so tests and such won't fail.
I don't want to make this function return error and change all calls to it just so in theory, maybe, someday, someone might do something terribly wrong. I think it is better to let Go panic in such a case.
I will change it to uint and check only x==0.

controllers/edgedeployment_controller_test.go
ydayagi: I made it exportable only temporarily. It should be **s**plitEdgeDevices.
I apologize for the confusion.

controllers/edgedeployment_controller.go
ydayagi: These are different functions and different inputs. I do not see how this can be done without making the code much more complex. Then the loop becomes redundant compared to extra complexity/code.
If you have an idea please suggest.
BTW, does it bother us that much? I mean iterating over the functions? Especially, when we have 100s of EdgeDevices?

controllers/edgedeployment_controller.go
ydayagi: It should be **s**plitEdgeDevices. I did not find a way to make it work with interface{}. All examples and docs I saw do not offer a solution.
If you have one please suggest.

controllers/edgedeployment_controller.go
pkliczewski: I would see this something similar to:

```
	inputs := SplitEdgeDevices(edgeDevices, r.Concurrency)
	var s []error

	var wg sync.WaitGroup
	wg.Add(len(inputs))
	
	for _, input := range inputs {
		go func() {
			defer wg.Done()
			err := fn(ctx, name, input)
			if err != nil {
				s = append(s, err)
			}
		}()
	}
	wg.Wait()
	if len(s) != 0 {
		return fmt.Errorf(mergeErrorMessages(s))
	}
	return nil
```

The only issue is to figure out common signature for `fn` if possible to make it more generic. Let me know what you think.

ydayagi: I would have done that if it was always the same function.
However, in our case, we have 3 functions and may have more.
So for each function, you need a loop that knows its input and its signature.
I don't think we can work around that. You will always have to create something specific to each function call and will always have to loop over all function calls.
Regarding the channel. I think it is much easier in our case to use a slice.
Our input is known in advance. So we can put it in a collection (e.g. slice).
A channel is an implementation of a queue (for our needs).
It is needed if you don't know when and how many inputs you'll have.
I know that it seems cool to use channels and mocks and a lot of stuff that many people use.
But, it does not mean that it's right.
Your call.

ydayagi: I googled 'go generics' and it seems that it's going to be included soon. This is what you're looking for.

ydayagi: we can do something like what you suggested if we limit executeConcurrent only for functions that handle []EdgeDevice and nothing else changes from call to call.

Makefile
```diff
@@ -101,7 +101,7 @@ vet: ## Run go vet against code.
 ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 test: manifests generate fmt vet test-fast ## Run tests.
 
-TEST_PACKAGES := ./...
+TEST_PACKAGES := ./controllers
```
masayag, line 104: why there is a need to narrow the packages ?

config/manager/controller_manager_config.properties
```diff
@@ -3,4 +3,5 @@ METRICS_ADDR=:8080
 PROBE_ADDR=:8081
 LEADER_ELECT=false
 WEBHOOK_PORT=9443
-OBC_AUTO_CREATE=true
\ No newline at end of file
+OBC_AUTO_CREATE=true
+EDGEDEPLOYMENT_CONCURRENCY=5
```
masayag, line 7: I'd avoid exposing it, there is a default of 5 internally and exposing it might give the user the impression that this can be tweaked as needed.

Makefile
ydayagi: oops....fixing :-)

config/manager/controller_manager_config.properties
ydayagi: Piotr asked for it. So I think you 2 need to talk about it.
My personal opinion is that it makes no difference.
No one changes things without checking what it means. Especially if they don't know what it is.
If I remove this line and the configMap won't have this field the code will still work, correct?
So actually, we're talking about this file only, correct?

masayag: yes, it will work, the value will be taken from its default in code.

controllers/edgedeployment_controller.go
```diff
@@ -285,3 +342,26 @@ func mergeErrorMessages(errs []error) string {
 	}
 	return message
 }
+
+func splitEdgeDevices(s []managementv1alpha1.EdgeDevice, x uint) [][]managementv1alpha1.EdgeDevice {
```
jakub-dzon, line 346: `s` and `x` names do not say anything; please change to something meaningful.

controllers/edgedeployment_controller.go
jakub-dzon: But will it panic? From what I see it shouldn't; at line https://github.com/jakub-dzon/k4e-operator/pull/38/files#diff-0d2d8c9ff48a926eddf2ba67be4f7dc0635e7b88376946057eba6dabfb0931ecR280 it will return nil immediately, which will return nil-error array, which would be handled gracefully. Nobody will notice the misconfiguration.
Maybe it should be validated in main and cause the app to crash right away?

controllers/edgedeployment_controller.go
```diff
@@ -236,13 +259,47 @@ func (r *EdgeDeploymentReconciler) getLabelledEdgeDevices(ctx context.Context, n
...
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *EdgeDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&managementv1alpha1.EdgeDeployment{}).
 		Complete(r)
 }
 
+func ConcurrentExecutor(concurrency uint, f ConcurrentFunc, input []managementv1alpha1.EdgeDevice) []error {
```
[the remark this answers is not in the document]
jakub-dzon: Please change `input` to something meaningful, like `edgeDevices`.

ll
```diff
@@ -0,0 +1,1215 @@
+/home/ydayagi/work/go/src/github.com/ydayagi/k4e-operator/bin/controller-gen "crd:trivialVersions=true,preserveUnknownFields=false" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
jakub-dzon, line 1: I think that file has gotten into the PR by accident.

jakub-dzon (changes requested):

controllers/edgedeployment_controller.go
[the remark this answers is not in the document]
ydayagi: I will change input to edgeDevices in order to save the debate. Although, I do not agree that it makes a difference or that it is wrong..
I did not understand what is wrong with the name 'ConcurrentExecutor'. The name describes exactly what the function is supposed to do.

controllers/edgedeployment_controller.go
ydayagi: please suggest names

controllers/edgedeployment_controller.go
ydayagi: it won't panic because you asked for a change. I would have left it as it was originally.
I think there's no need to handle every theoretical problem. I can not imagine someone setting the value to 0 on purpose.
and even someone did, the effect would be immediate (e.g. no deployments start/stop in devices).
But again, for the sake of saving time debating I will add validation in main.

controllers/edgedeployment_controller.go
jakub-dzon: edgeDevices and splitSize?

controllers/edgedeployment_controller.go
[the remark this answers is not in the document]
jakub-dzon: It's a noun; function by definition is not representing some object/thing (noun), but action (verb). I'd accept a noun name of a function for an accessor or a constructor. It's neither.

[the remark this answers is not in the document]
ydayagi: That is very much arguable and I do not agree with this approach.

ll
ydayagi: thanks

controllers/edgedeployment_controller.go
ydayagi: so splitSize is the size of each slice after the split? this name is confusing. Anyone who reads the function understands immediately what x and s represent. This is why I left it s and x. And if we really want to get philosophical then the type of the slice is meaningless here. Therefore using a name that relates to the type (edgeDevices) is not appropriate in my opinion.
But again, I will change it just to avoid wasting our time in debate.

controllers/edgedeployment_controller.go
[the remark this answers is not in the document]
pkliczewski: In general there are naming conventions which should be followed. Here are [golang](https://go.dev/doc/effective_go#names) naming conventions. We can come up with our own if we want to adjust generic golang conventions. It is important to have consistent code in the project.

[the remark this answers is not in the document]
ydayagi: I did not find something in golang's conventions that contradicts the name I used. I use golint and other go tools that help me follow the conventions.
I think that the debate here is not about conventions but personal flavor.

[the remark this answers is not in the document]
pkliczewski: Let's talk about this. As far as I can see nothing personal here.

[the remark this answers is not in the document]
ydayagi: @jakub-dzon I will change it to ExecuteConcurrent. Please add explanations to contributing guidelines so it will be easier to solve these issues in the future.

jakub-dzon (approved):

masayag (approved):

main.go
```diff
@@ -128,7 +137,10 @@ func main() {
 	setupLog = ctrl.Log
 	setupLog.Info("Started with configuration", "configuration", Config)
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	r, err := ctrl.GetConfig()
+	r.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(100, 1000)
+	mgr, err := ctrl.NewManager(r, ctrl.Options{
+		//mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```
masayag, line 143: actually, merged too quickly - I missed this commented line. @ydayagi pls remove it.

PR #43, opened 2021-11-25
Provide custom image repositories configuration to edge devices

jakub-dzon:
This PR adds code that is responsible for retrieving authfile secret and providing its content to the edge devices.


pkliczewski (commented):
What should we do when we update a edgedeployment with new version of container but the secret is not available anymore? Maybe we should keep old version running on the device?

jakub-dzon:
@pkliczewski - updated the code to return error when secret is missing. For user visibility an event is recorded.

pkliczewski (commented):
/lgtm

internal/images/repo-auth.go
```diff
@@ -0,0 +1,39 @@
...
+	secret := v1.Secret{}
+	err := r.client.Get(ctx, client.ObjectKey{Namespace: namespace, Name: name}, &secret)
+	if err != nil {
+		return "", err
+	}
+	authFile, found := secret.Data[".dockerconfigjson"]
+	if !found {
+		return "", errors.New(fmt.Sprintf(".dockerconfigjson not found in %s/%s Secret", namespace, name))
+	}
+	if err != nil {
```
masayag, line 34: this if is redundant

internal/yggdrasil/yggdrasil.go
```diff
@@ -43,15 +44,18 @@ type Handler struct {
...
-func NewYggdrasilHandler(deviceRepository edgedevice.Repository, deploymentRepository edgedeployment.Repository, claimer *storage.Claimer, initialNamespace string, recorder record.EventRecorder) *Handler {
+func NewYggdrasilHandler(deviceRepository edgedevice.Repository, deploymentRepository edgedeployment.Repository,
+	claimer *storage.Claimer, initialNamespace string, recorder record.EventRecorder, repositoryAuth images.RepositoryAuthAPI) *Handler {
 	return &Handler{
 		deviceRepository:     deviceRepository,
 		deploymentRepository: deploymentRepository,
 		claimer:              claimer,
 		initialNamespace:     initialNamespace,
 		recorder:             recorder,
+		repositoryAuth:       repositoryAuth,
```
masayag, line 58: a question about the naming, in terms of symmetry with `deviceRepository` and `deploymentRepository`, shouldn't it be called `AuthRepository` ?

internal/yggdrasil/yggdrasil.go
```diff
@@ -310,33 +317,66 @@ func (h *Handler) updateDeploymentStatuses(oldDeployments []v1alpha1.Deployment,
...
 			}
 			data = &models.DataConfiguration{Paths: paths}
 		}
+
 		workload := models.Workload{
 			Name:          deployment.Name,
 			Specification: string(podSpec),
 			Data:          data,
 		}
+		authFile, err := h.getAuthFile(ctx, spec.ImageRepositories, deployment.Namespace)
```
masayag, line 346: I'm having second thoughts about the naming, ImageRepositories vs ImageRegistries.
Short quiz ;-) : The auth file should allow the pod to pull its image from the image _[complete the missing word]_.

In the context of images and k8s, it seems that the term `registry` is commonly used for it. https://kubernetes.io/docs/concepts/containers/images/

To the user it should be clear what is the semantic of repository, but if it is a synonym to the registry, we should pick the latter. WDYT ?

internal/yggdrasil/yggdrasil.go
```diff
@@ -310,33 +317,66 @@ func (h *Handler) updateDeploymentStatuses(oldDeployments []v1alpha1.Deployment,
...
-	return list
+	return list, nil
+}
+
+func (h *Handler) getAuthFile(ctx context.Context, imageRepositories *v1alpha1.ImageRepositoriesConfiguration, defaultNamespace string) (string, error) {
+	if imageRepositories != nil {
+		if secretRef := imageRepositories.AuthFileSecret; secretRef != nil {
+			namespace := secretRef.Namespace
+			if secretRef.Namespace == "" {
+				namespace = defaultNamespace
```
masayag, line 369: Shouldn't we prefer the namespace of the edge-device if it wasn't explicitly provided ?

masayag (changes requested):

internal/images/repo-auth.go
jakub-dzon: Thanks! How did it get there?;)

internal/yggdrasil/yggdrasil.go
jakub-dzon: If kept the naming in line, that would be RepositoryAuthRepository; It's not a repository for auth, but (image) respository authentication.

internal/yggdrasil/yggdrasil.go
jakub-dzon: It would be probably clearer. The downside is that the data model would have to be changed for consistency.
WDYT about changing that in this PR?

internal/yggdrasil/yggdrasil.go
jakub-dzon: `defaultNamespace` here comes from the parameter, which in turn is set with `deployment.Namespace` value, which is equal to device namespace.

internal/yggdrasil/yggdrasil.go
masayag: +1

internal/yggdrasil/yggdrasil.go
masayag: Right...I somehow assumed it is a constant defined as 'default' and missed the fact it was provided as a parameter :-)

internal/images/registry-auth_integration_test.go
```diff
@@ -0,0 +1,78 @@
...
+        "https://index.docker.io/v1/": {
+            "auth": "dGVzdC10ZXN0Cg=="
+        }
+    }
+}`
+)
+
+var _ = Describe("Images", func() {
+	var (
+		repoAuth *images.RegistryAuth
```
masayag, line 24: nit, s/repoAuth/registryAuth

internal/yggdrasil/yggdrasil.go
```diff
@@ -310,33 +317,66 @@ func (h *Handler) updateDeploymentStatuses(oldDeployments []v1alpha1.Deployment,
...
+				AuthFile: authFile,
+			}
+		}
 		list = append(list, &workload)
 	}
-	return list
+	return list, nil
+}
+
+func (h *Handler) getAuthFile(ctx context.Context, imageRepositories *v1alpha1.ImageRegistriesConfiguration, defaultNamespace string) (string, error) {
```
masayag, line 364: s/imageRepositories/imageRegistries ?

internal/yggdrasil/yggdrasil_test.go
```diff
@@ -432,7 +434,156 @@ var _ = Describe("Yggdrasil", func() {
 
 			Expect(config.DeviceID).To(Equal(deviceName))
 			Expect(config.Workloads).To(HaveLen(1))
-			Expect(config.Workloads[0].Name).To(Equal("workload1"))
+			workload := config.Workloads[0]
+			Expect(workload.Name).To(Equal("workload1"))
+			Expect(workload.ImageRegistries).To(BeNil())
+		})
+
+		It("Image repository authfile is included", func() {
```
masayag, line 442: s/repository/registry (several occurrences in this file)

internal/yggdrasil/yggdrasil.go
```diff
@@ -310,33 +317,66 @@ func (h *Handler) updateDeploymentStatuses(oldDeployments []v1alpha1.Deployment,
...
 			Name:          deployment.Name,
 			Specification: string(podSpec),
 			Data:          data,
 		}
+		authFile, err := h.getAuthFile(ctx, spec.ImageRegistries, deployment.Namespace)
+		if err != nil {
+			h.recorder.Eventf(device, corev1.EventTypeWarning, "Misconfiguration",
+				"Auth file secret %s used by deployment %s/%s is missing", spec.ImageRegistries.AuthFileSecret.Name, deployment.Namespace, deployment.Name,
+			)
+			logger.Error(err, "Can't get auth file")
```
masayag, line 351: pls add also details about the missing secret/deployment to the log.

internal/yggdrasil/yggdrasil.go
```diff
@@ -310,33 +317,66 @@ func (h *Handler) updateDeploymentStatuses(oldDeployments []v1alpha1.Deployment,
...
 		}
+
 		workload := models.Workload{
 			Name:          deployment.Name,
 			Specification: string(podSpec),
 			Data:          data,
 		}
+		authFile, err := h.getAuthFile(ctx, spec.ImageRegistries, deployment.Namespace)
+		if err != nil {
+			h.recorder.Eventf(device, corev1.EventTypeWarning, "Misconfiguration",
```
masayag, line 348: this one is tricky :-) the event will be reported for the device, but in fact, the misconfiguration is on the edgeployment.
this might end with many events for the devices that will try to run the misconfigured deployment.
in terms of context and amount of potentials event, perhaps edgedeployment will be a better fit ?

masayag (changes requested):

internal/yggdrasil/yggdrasil.go
jakub-dzon: I thought about it a bit and decided to go with the edge device, because that was also the result of more general discussion around events before Ondra's changes. User adding a deployment right now should monitor edgedevice for updates about the deployment on device status and they would also see the events in there; there is no reason to monitor the edgedeployment and adding events there might be confusing at this point.
If we decide to refactor edgedeployment - edgedevice relationship, then maybe at that point we should refactor the events as well.

masayag (approved):

PR #44, opened 2021-11-26
TLS: Enable Mutual TLS connection

eloycoto:
This commit introduces the following changes:

TLS Listening
==============

When the operator starts, it will try to search for a CA certificate on
the 'k4e-ca' secret, where it will retrieve the x509 cert and the RSA
key. It'll create a new ca-cert if it's not present.

The operator will create the server certificates with that CA cert and
start listening on port 8443 by default. DNS name can be retrieval via
config; by default is k4e.com.

To keep compatibility with the device-worker, the server stills listen
on the HTTP port.

Mutual TLS
===========

To use the API, now client certificates are required, and as defined in
the task, there are two kinds of scenarios:

- **Registered endpoint**: Where a client certificate must be used, if
  the cert is expired, it'll be able to register the device. We only
  check the given CA signed the client certificate.

- **Auth endpoint**: Where a valid client certificate is needed.

This is why certs are checked on each request.

What is missing:
----------------

- The revocation list, a new CR need to be created and check on
  `mtls.VerifyRequest` ECOPROJECT-218
- Register endpoint is not yet implemented, where need to sign given
  x509 certificates. ECOPROJECT-401
- AuthN is done, but Authz on yggdrasil.API need to be done, like
  deviceID matchs with certificate CN. ECOPROJECT-402
- Doc

Coverage:
----------
Coverage is in good shape, and I'm confident with it:

```
github.com/jakub-dzon/k4e-operator/internal/mtls/ca.go:44:			NewMTLSconfig			100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/ca.go:61:			SetCAProvider			100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/ca.go:65:			InitCertificates		88.5%
github.com/jakub-dzon/k4e-operator/internal/mtls/ca.go:118:			CreateRegistrationClient	87.5%
github.com/jakub-dzon/k4e-operator/internal/mtls/ca.go:149:			isClientCertificateSigned	100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/ca.go:171:			VerifyRequest			100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/caprovider_secret.go:36:	NewCASecretProvider		100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/caprovider_secret.go:43:	GetName				100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/caprovider_secret.go:47:	GetCACertificate		86.7%
github.com/jakub-dzon/k4e-operator/internal/mtls/caprovider_secret.go:89:	CreateRegistrationCertificate	80.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/cert_utils.go:27:		NewCertificateGroupFromCACM	76.5%
github.com/jakub-dzon/k4e-operator/internal/mtls/cert_utils.go:58:		CreatePem			100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/cert_utils.go:76:		parseSignedCertificate		100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/cert_utils.go:83:		GetCertificate			100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/cert_utils.go:87:		GetCert				100.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/cert_utils.go:91:		getCACertificate		81.8%
github.com/jakub-dzon/k4e-operator/internal/mtls/cert_utils.go:127:		getKeyAndCSR			80.0%
github.com/jakub-dzon/k4e-operator/internal/mtls/cert_utils.go:151:		getServerCertificate		100.0%
```

Fix https://issues.redhat.com/browse/ECOPROJECT-400

### DIFF
--- a/internal/mtls/ca.go
+++ b/internal/mtls/ca.go
@@ -21,7 +21,7 @@ const (
 	regClientSecretLabelKey      = "reg-client-ca"
 )
 
-// The main reason to have an interface here is to be able to extend this to
+// CAProvider The main reason to have an interface here is to be able to extend this to
 // future Cert providers, like:
 // - Vault
 // - Acme protocol
@@ -41,7 +41,7 @@ type TLSConfig struct {
 	namespace        string
 }
 
-func NewMTLSconfig(client client.Client, namespace string, domains []string, localhostEnabled bool) *TLSConfig {
+func NewMTLSConfig(client client.Client, namespace string, domains []string, localhostEnabled bool) *TLSConfig {
 	config := &TLSConfig{
 		config:           nil,
 		client:           client,
@@ -64,7 +64,7 @@ func (conf *TLSConfig) SetCAProvider(caProviders []CAProvider) {
 
 func (conf *TLSConfig) InitCertificates() (*tls.Config, []*x509.Certificate, error) {
 	if len(conf.caProvider) == 0 {
-		return nil, nil, fmt.Errorf("No provider set")
+		return nil, nil, fmt.Errorf("no CA provider is set")
 	}
 
 	var errors error
@@ -92,7 +92,7 @@ func (conf *TLSConfig) InitCertificates() (*tls.Config, []*x509.Certificate, err
 	}
 
 	if len(caCerts) == 0 {
-		return nil, nil, fmt.Errorf("Cannot get any CA certificate")
+		return nil, nil, fmt.Errorf("cannot get any CA certificate")
 	}
 
 	// We always sign the certificates with the first CA server. I guess that it's normal
@@ -148,20 +148,15 @@ func (conf *TLSConfig) CreateRegistrationClientCerts() error {
 // for registration endpoint, we cannot assume that it'll be ok.
 func isClientCertificateSigned(PeerCertificates []*x509.Certificate, CAChain []*x509.Certificate) bool {
 	for _, cert := range PeerCertificates {
-		certValid := false
 		for _, caCert := range CAChain {
 			err := cert.CheckSignatureFrom(caCert)
 			// TODO log debug here with the error. Can be too verbose.
 			if err == nil {
-				certValid = true
-				break
+				return true
 			}
 		}
-		if !certValid {
-			return false
-		}
 	}
-	return true
+	return false
 }
 
 // VerifyRequest check certificate based on the scenario needed:
@@ -179,15 +174,14 @@ func VerifyRequest(r *http.Request, verifyType int, verifyOpts x509.VerifyOption
 		return res
 	}
 
-	valid := true
 	for _, cert := range r.TLS.PeerCertificates {
 		if cert.Subject.CommonName == certRegisterCN {
-			valid = false
+			return false
 		}
 		if _, err := cert.Verify(verifyOpts); err != nil {
 			// TODO log debug here with the error. Can be too verbose.
 			return false
 		}
 	}
-	return valid
+	return true
 }
--- a/internal/mtls/ca.go
+++ b/internal/mtls/ca.go
@@ -115,15 +115,15 @@ func (conf *TLSConfig) InitCertificates() (*tls.Config, []*x509.Certificate, err
 	return tlsConfig, CACertChain, nil
 }
 
-func (conf *TLSConfig) CreateRegistrationClient() error {
-
-	name := fmt.Sprintf("%s-%s",
-		regClientSecretNamePrefix,
-		utilrand.String(regClientSecretNameRandomLen))
+func (conf *TLSConfig) CreateRegistrationClientCerts() error {
 
 	if len(conf.caProvider) == 0 {
 		return fmt.Errorf("Cannot get ca provider")
 	}
+
+	name := fmt.Sprintf("%s-%s",
+		regClientSecretNamePrefix,
+		utilrand.String(regClientSecretNameRandomLen))
 
 	certData, err := conf.caProvider[0].CreateRegistrationCertificate(name)
 	if err != nil {
--- a/internal/mtls/ca.go
+++ b/internal/mtls/ca.go
@@ -1,0 +1,193 @@
+package mtls
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net/http"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/jakub-dzon/k4e-operator/internal/yggdrasil"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	regClientSecretNameRandomLen = 10
+	regClientSecretNamePrefix    = "reg-client-ca"
+	regClientSecretLabelKey      = "reg-client-ca"
+)
+
+// The main reason to have an interface here is to be able to extend this to
+// future Cert providers, like:
+// - Vault
+// - Acme protocol
+// Keeping as an interface, so in future users can decice.
+type CAProvider interface {
+	GetName() string
+	GetCACertificate() (*CertificateGroup, error)
+	CreateRegistrationCertificate(name string) (map[string][]byte, error)
+}
+
+type TLSConfig struct {
+	config           *tls.Config
+	client           client.Client
+	caProvider       []CAProvider
+	Domains          []string
+	LocalhostEnabled bool
+	namespace        string
+}
+
+func NewMTLSconfig(client client.Client, namespace string, domains []string, localhostEnabled bool) *TLSConfig {
+	config := &TLSConfig{
+		config:           nil,
+		client:           client,
+		Domains:          domains,
+		namespace:        namespace,
+		LocalhostEnabled: localhostEnabled,
+	}
+
+	// Secret providers here
+	secretProvider := NewCASecretProvider(client, namespace)
+	config.caProvider = append(config.caProvider, secretProvider)
+
+	return config
+}
+
+// @TODO mainly used for testing, maybe not needed at all
+func (conf *TLSConfig) SetCAProvider(caProviders []CAProvider) {
+	conf.caProvider = caProviders
+}
+
+func (conf *TLSConfig) InitCertificates() (*tls.Config, []*x509.Certificate, error) {
+	if len(conf.caProvider) == 0 {
+		return nil, nil, fmt.Errorf("No provider set")
+	}
+
+	var errors error
+	caCerts := []*CertificateGroup{}
+
+	CACertChain := []*x509.Certificate{}
+	caCertPool := x509.NewCertPool()
+
+	for _, caProvider := range conf.caProvider {
+		caCert, err := caProvider.GetCACertificate()
+		if err != nil {
+			errors = multierror.Append(errors, fmt.Errorf(
+				"cannot get CA certificate for provider %s: %v",
+				caProvider.GetName(), err))
+			continue
+		}
+
+		caCerts = append(caCerts, caCert)
+		CACertChain = append(CACertChain, caCert.GetCert())
+		caCertPool.AppendCertsFromPEM(caCert.certPEM.Bytes())
+	}
+
+	if errors != nil {
+		return nil, nil, errors
+	}
+
+	if len(caCerts) == 0 {
+		return nil, nil, fmt.Errorf("Cannot get any CA certificate")
+	}
+
+	// We always sign the certificates with the first CA server. I guess that it's normal
+	serverCert, err := getServerCertificate(conf.Domains, conf.LocalhostEnabled, caCerts[0])
+	if err != nil {
+		return nil, nil, err
+	}
+
+	certificate, err := serverCert.GetCertificate()
+	if err != nil {
+		return nil, nil, fmt.Errorf("Cannot create server certfificate: %v", err)
+	}
+
+	tlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{certificate},
+		ClientCAs:    caCertPool,
+		ClientAuth:   tls.RequireAnyClientCert,
+		MinVersion:   tls.VersionTLS13,
+	}
+	return tlsConfig, CACertChain, nil
+}
+
+func (conf *TLSConfig) CreateRegistrationClient() error {
+
+	name := fmt.Sprintf("%s-%s",
+		regClientSecretNamePrefix,
+		utilrand.String(regClientSecretNameRandomLen))
+
+	if len(conf.caProvider) == 0 {
+		return fmt.Errorf("Cannot get ca provider")
+	}
+
+	certData, err := conf.caProvider[0].CreateRegistrationCertificate(name)
+	if err != nil {
+		return fmt.Errorf("Cannot create client certificate")
+	}
+
+	secret := corev1.Secret{
+		ObjectMeta: v1.ObjectMeta{
+			Namespace: conf.namespace,
+			Name:      name,
+			Labels:    map[string]string{regClientSecretLabelKey: "true"},
+		},
+		Data: certData,
+	}
+
+	return conf.client.Create(context.TODO(), &secret)
+}
+
+// isClientCertificateSigned is checking that PeerCertificates are signed by at
+// least one give CA certificate. The main reason to do this and not
+// x509.Certificate.Cert(certStore) is because is checking expiration time, and
+// for registration endpoint, we cannot assume that it'll be ok.
+func isClientCertificateSigned(PeerCertificates []*x509.Certificate, CAChain []*x509.Certificate) bool {
+	for _, cert := range PeerCertificates {
+		certValid := false
+		for _, caCert := range CAChain {
+			err := cert.CheckSignatureFrom(caCert)
+			// TODO log debug here with the error. Can be too verbose.
+			if err == nil {
+				certValid = true
+				break
+			}
+		}
+		if !certValid {
+			return false
+		}
+	}
+	return true
+}
+
+// VerifyRequest check certificate based on the scenario needed:
+// registration endpoint: Any cert signed, even if it's expired.
+// All endpoints: checking that it's valid certificate.
+// @TODO check here the list of rejected certificates.
+func VerifyRequest(r *http.Request, verifyType int, verifyOpts x509.VerifyOptions, CACertChain []*x509.Certificate) bool {
+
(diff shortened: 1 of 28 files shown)


main.go
```diff
@@ -56,12 +58,28 @@ const (
...
 )
 
 var Config struct {
+
+	// NOTE DEPRECATE
 	// The port of the HTTP server
 	HttpPort uint16 `envconfig:"HTTP_PORT" default:"8888"`
 
+	// The port of the HTTPs server
+	HttpsPort uint16 `envconfig:"HTTP_PORT" default:"8443"`
```
jakub-dzon, line 74: It needs to be exposed in the service (https://github.com/jakub-dzon/k4e-operator/blob/e982e6d7a61045ccab22161cf48426da803953dc/config/manager/manager.yaml#L60).

internal/mtls/cert_utils.go
```diff
@@ -0,0 +1,174 @@
...
+		SerialNumber: CACert.cert.SerialNumber,
+		Subject: pkix.Name{
+			CommonName:   "*", // CommonName match all, and using ASN names
+			Organization: []string{serverCertOrganization},
+			Country:      []string{"US"},
+		},
+		DNSNames:     dnsNames,
+		IPAddresses:  ips,
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().AddDate(certDefaultExpiration, 0, 0),
```
jakub-dzon, line 168: Do I understand correctly that every time we start the operator, the server certificate expiration changes? Meaning - it would have the same serial number and other details but different expiration every time?

internal/yggdrasil/yggdrasil.go
```diff
@@ -55,6 +62,24 @@ func NewYggdrasilHandler(deviceRepository edgedevice.Repository, deploymentRepos
 	}
 }
 
+func isRegisteredURL(url *url.URL) bool {
```
jakub-dzon, line 65: `isRegistrationURL`?

internal/yggdrasil/yggdrasil.go
```diff
@@ -55,6 +62,24 @@ func NewYggdrasilHandler(deviceRepository edgedevice.Repository, deploymentRepos
...
+	parts := strings.Split(url.Path, "/")
+	if len(parts) == 0 {
+		return false
+	}
+
+	last := parts[len(parts)-1]
+	return last == "registration"
+}
+
+func (h *Handler) SetAuthType(r *http.Request) int {
```
jakub-dzon, line 75: WDYT about `GetAuthType` instead? And maybe moving it to a separate file? This method is not part of the HTTP API after all.

internal/mtls/ca_test.go
```diff
@@ -0,0 +1,508 @@
...
+			err = k8sClient.Create(context.TODO(), &nsSpec)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		AfterEach(func() {
+			err := testEnv.Stop()
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("No namespace does not exists", func() {
```
jakub-dzon, line 73: nit: "No namespace exists"

internal/mtls/ca_test.go
```diff
@@ -0,0 +1,508 @@
...
+				}
+
+				// when
+				res := mtls.VerifyRequest(r, AuthType, opts, CAChain)
+
+				// then
+				Expect(res).To(BeFalse())
+			})
+
+			It("Lastet CA certificate is valid", func() {
```
jakub-dzon, line 280: nit: `Latest` or `Last`?

internal/mtls/ca_test.go
```diff
@@ -0,0 +1,508 @@
...
+				}
+
+				// when
+				res := mtls.VerifyRequest(r, AuthType, opts, CAChain)
+
+				// then
+				Expect(res).To(BeTrue())
+			})
+
+			It("Invalid certificate is correct", func() {
```
jakub-dzon, line 363: "... is incorrect"?

jakub-dzon (commented):
Great work! Have couple of doubts and suggestions.

internal/mtls/cert_utils.go
eloycoto: at some point, we're going to change this logic to retrieve the cert from a 3party provider, but It's the easiest way that I found right now.

internal/mtls/ca.go
```diff
@@ -0,0 +1,193 @@
...
+	return tlsConfig, CACertChain, nil
+}
+
+func (conf *TLSConfig) CreateRegistrationClient() error {
+
+	name := fmt.Sprintf("%s-%s",
+		regClientSecretNamePrefix,
+		utilrand.String(regClientSecretNameRandomLen))
+
+	if len(conf.caProvider) == 0 {
```
masayag, line 124: this can be the first condition in this function, before defining the `name`

internal/mtls/cert_utils.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func getCACertificate() (*CertificateGroup, error) {
+	ca := &x509.Certificate{
+		SerialNumber: big.NewInt(time.Now().Unix()),
+		Subject: pkix.Name{
+			Organization: []string{"K4e-operator"},
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(10, 0, 0),
```
masayag, line 98: Isn't 10 years validity for the certificate too long ? Should the certificate period be configurable and set based on org policy ?

internal/mtls/cert_utils.go
```diff
@@ -0,0 +1,174 @@
...
+	certGroup.signedCert = ca
+	certGroup.privKey = key
+	return certGroup, nil
+}
+
+// CreatePem from the load certificates create the PEM file and stores in local
+func (c *CertificateGroup) CreatePem() {
+
+	caPEM := new(bytes.Buffer)
+	pem.Encode(caPEM, &pem.Block{
```
masayag, line 61: no need to consider potential error ?

internal/mtls/cert_utils.go
```diff
@@ -0,0 +1,174 @@
...
+		Type:  "CERTIFICATE",
+		Bytes: c.certBytes,
+	})
+
+	c.certPEM = caPEM
+
+	privKeyPEM := new(bytes.Buffer)
+	pem.Encode(privKeyPEM, &pem.Block{
```
masayag, line 69: no need to consider potential error ?

internal/mtls/cert_utils.go
```diff
@@ -0,0 +1,174 @@
...
+		cert:      ca,
+		privKey:   caPrivKey,
+		certBytes: caBytes,
+	}
+	certificateBundle.CreatePem()
+	certificateBundle.parseSignedCertificate()
+	return &certificateBundle, nil
+}
+
+func getKeyAndCSR(cert *x509.Certificate, caCert *CertificateGroup) (*CertificateGroup, error) {
```
masayag, line 127: perhaps replace `get` with `create` to reflect new cert is created by each call.

internal/mtls/caprovider_secret.go
```diff
@@ -0,0 +1,121 @@
...
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().AddDate(certDefaultExpiration, 0, 0),
+		SubjectKeyId: []byte{1, 2, 3, 4, 6},
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		KeyUsage:     x509.KeyUsageDigitalSignature,
+	}
+
+	certGroup, err := getKeyAndCSR(cert, CACert)
+	if err != nil {
+		return nil, fmt.Errorf("Cannot sign certificate request")
```
masayag, line 112: Will including the `err` in the error message provide more info to the user about the failure ?

main.go
masayag: 8443 is taken by by the other rbac proxy container
```
→ git grep 8443
config/default/manager_auth_proxy_patch.yaml:        - "--secure-listen-address=0.0.0.0:8443"
config/default/manager_auth_proxy_patch.yaml:        - containerPort: 8443
config/rbac/auth_proxy_service.yaml:    port: 8443

```

main.go
```diff
@@ -154,16 +172,74 @@ func main() {
 	}
 
 	go func() {
+
```
masayag, line 175: how this should behave when we have more than a single replica ?
shouldn't it be invoked only by the leader ?

main.go
```diff
@@ -154,16 +172,74 @@ func main() {
...
+			YggdrasilAPI: yggdrasilAPIHandler,
+			InnerMiddleware: func(h http.Handler) http.Handler {
+				// This is needed for one reason. Registration endpoint can be
+				// triggered with a certificate signed by the CA, but can be expired
+				// The main reason to allow expired certificates in this endpoint, it's
+				// to renew client certificates, and because some devices can be
+				// disconnected for days and does not have the option to renew it.
+				return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					authType := yggdrasilAPIHandler.SetAuthType(r)
+					if !mtls.VerifyRequest(r, authType, opts, CACertChain) {
```
masayag, line 219: Will this be affected by exposing a service as a route and getting a different URL for yggdrasil server ?
 (by `oc expose svc k4e-operator-controller-manager -n k4e-operator-system`).

```
→ oc get routes -n k4e-operator-system k4e-operator-controller-manager
NAME                              HOST/PORT                                                                       PATH   SERVICES                          PORT   TERMINATION   WILDCARD
k4e-operator-controller-manager   k4e-operator-controller-manager-k4e-operator-system.apps.mycluster.redhat.com          k4e-operator-controller-manager   http                 None
```

internal/mtls/cert_utils.go
eloycoto: This is only for CA

internal/yggdrasil/yggdrasil.go
eloycoto: Keep here until ECOPROJECT-401 check this again.

main.go
eloycoto: No, not really, I mean, it'll fetch the config from the configmap or 3party secret provider.

main.go
eloycoto: No, that's the part of the server, what we're checking here is the client certificates. So no an issue at all.

internal/yggdrasil/yggdrasil.go
eloycoto: To be checked on ECOPROJECT-401

internal/yggdrasil/yggdrasil.go
jakub-dzon: I was suggesting changing the name of the method to `isRegistrationURL`; after all it's not a "registered URL".

internal/yggdrasil/yggdrasil.go
jakub-dzon: What do you mean by that?

internal/mtls/cert_utils.go
pkliczewski: How do we rotate CA when 10y passes? We can check validation period and regenerate and we would need to force agents to reregister.

internal/mtls/caprovider_secret.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Cannot retrieve caCert")
+	}
+
+	cert := &x509.Certificate{
+		SerialNumber: CACert.cert.SerialNumber,
+		Subject: pkix.Name{
+			CommonName:   certRegisterCN,
+			Organization: []string{certOrganization},
+			// Country:      []string{"US"},
```
pkliczewski, line 100: I think we can remove the country

internal/mtls/caprovider_secret.go
```diff
@@ -0,0 +1,121 @@
...
+	providerName = "secret"
+
+	caCertSecretKey     = "ca.key"
+	caCertCertKey       = "ca.crt"
+	clientCertSecretKey = "client.key"
+	clientCertCertKey   = "client.crt"
+
+	certOrganization       = "k4e-operator"
+	certRegisterCN         = "register"
+	certDefaultExpiration  = 1 // years
```
pkliczewski, line 27: Let's make this configurable.

internal/mtls/ca.go
```diff
@@ -0,0 +1,193 @@
...
+
+	if verifyType == yggdrasil.YggdrasilRegisterAuth {
+		res := isClientCertificateSigned(r.TLS.PeerCertificates, CACertChain)
+		return res
+	}
+
+	valid := true
+	for _, cert := range r.TLS.PeerCertificates {
+		if cert.Subject.CommonName == certRegisterCN {
+			valid = false
```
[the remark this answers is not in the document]
eloycoto: umm. no, client can send two certificates, and one of them can be valid.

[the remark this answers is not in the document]
pkliczewski: My intention was to simplify how we handle return value. It seems we do not need `valid` variable.

internal/mtls/ca.go
```diff
@@ -0,0 +1,193 @@
...
+
+	if len(conf.caProvider) == 0 {
+		return fmt.Errorf("Cannot get ca provider")
+	}
+
+	name := fmt.Sprintf("%s-%s",
+		regClientSecretNamePrefix,
+		utilrand.String(regClientSecretNameRandomLen))
+
+	certData, err := conf.caProvider[0].CreateRegistrationCertificate(name)
```
masayag, line 128: Can you explain how will this behave with multiple CA providers since in `InitCertificates` we iterates over all of them and here only the first one is picked.

masayag (approved):

internal/mtls/ca.go
```diff
@@ -0,0 +1,193 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	regClientSecretNameRandomLen = 10
+	regClientSecretNamePrefix    = "reg-client-ca"
+	regClientSecretLabelKey      = "reg-client-ca"
+)
+
+// The main reason to have an interface here is to be able to extend this to
```
masayag, line 24: documentation should start with `CAProvider`

masayag (changes requested):

internal/mtls/cert_utils.go
eloycoto: We can follow the same way as here: 
https://www.cockroachlabs.com/docs/stable/rotate-certificates.html

internal/mtls/cert_utils.go
```diff
@@ -0,0 +1,184 @@
...
+	signedCert *x509.Certificate
+	privKey    *rsa.PrivateKey
+	certBytes  []byte
+	certPEM    *bytes.Buffer
+	PrivKeyPEM *bytes.Buffer
+}
+
+func NewCertificateGroupFromCACM(configMap map[string][]byte) (*CertificateGroup, error) {
+	certGroup := &CertificateGroup{
+		certPEM:    bytes.NewBuffer(configMap["ca.crt"]),
```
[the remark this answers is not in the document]
eloycoto: yes, can be an issue, a big corner case, I do not think that it's possible. Added a @TODO and let's check in the near future., is that ok for you?

internal/mtls/ca.go
```diff
@@ -0,0 +1,187 @@
...
+		CACertChain = append(CACertChain, caCert.GetCert())
+		caCertPool.AppendCertsFromPEM(caCert.certPEM.Bytes())
+	}
+
+	if errors != nil {
+		return nil, nil, errors
+	}
+
+	if len(caCerts) == 0 {
+		return nil, nil, fmt.Errorf("Cannot get any CA certificate")
```
masayag, line 95: as a general comment for all `fmt.Errorf` calls, according to the guidelines:
"Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation"

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

internal/mtls/cert_utils.go
ydayagi: Do we create new cert and key upon restart even if it exists?

internal/mtls/ca.go
eloycoto: you cannot sign a certificate with multiple CA.

eloycoto:
Mark as WIP until got the TLS server certificate in a CM.

internal/mtls/caprovider_secret.go
eloycoto: Using ENV variable?

internal/mtls/cert_utils.go
eloycoto: Is this blocking the PR, should I implement the process to Update the CA cert?

internal/mtls/cert_utils.go
eloycoto: mark the PR as WIP and I'll implement the configmap retrieval for this

ydayagi: It seems we call InitCertificates during startup. Eventually, getServerCertificate is called. It creates a new CA cert. Is this correct? Don't we store the CA somewhere?

internal/mtls/caprovider_secret.go
masayag: yes, ENV variable is the option to go. we can just define it via ENV but not necessarily expose it in the configmap.

internal/mtls/cert_utils.go
eloycoto: CA cert is stored, TLS server certificate no.

ydayagi: TLS server certificate is the one used by the operator? I mean the HTTP server within the operator? and we create a new one after every startup?

eloycoto: Jira issue to track this: 
https://issues.redhat.com/browse/ECOPROJECT-431

internal/mtls/cert_utils.go
eloycoto: Issue to track this: 
https://issues.redhat.com/browse/ECOPROJECT-432

internal/mtls/caprovider_secret.go
eloycoto: https://issues.redhat.com/browse/ECOPROJECT-433

jakub-dzon (approved):

pkliczewski:
commend from @masayag on slack: "as long as we maintain the http for now, we can merge the PR"

PR #46, opened 2021-12-05
Adding new metrics

bardielle:


internal/metrics/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+			Help: "Number of registered EdgeDevices",
+		},
+	)
+	failedToCompleteRegistrationEdgeDevices = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: failedToCompleteRegistrationEdgeDevicesQuery,
+			Help: "Number of failed registration EdgeDevices",
+		},
+	)
+	createdEdgeDeployments = prometheus.NewGauge(
```
masayag, line 36: pls replace with https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Counter
it will replace the need to re-calculate the numbers.

pkliczewski:
@machacekondra please take a look

internal/metrics/metrics.go
machacekondra: Counter is used only for incrementing, the number of edgedeployments can be increased and decreased, isn't the `Gauge` the proper type, here?

config/manager/manager.yaml
```diff
@@ -62,6 +62,26 @@ spec:
...
+      targetPort: 8080
   selector:
     control-plane: controller-manager
   type: ClusterIP
+---
+kind: ServiceMonitor
+apiVersion: monitoring.coreos.com/v1
+metadata:
+  name: k4e-operator-servicemonitor
+  namespace: system
```
machacekondra, line 77: k4e-operator-system

config/manager/manager.yaml
```diff
@@ -62,6 +62,26 @@ spec:
...
+  labels:
```
machacekondra, line 78: Any reason to label the ServiceMonitor?

internal/metrics/metrics.go
masayag: Yes, you are correct. I missed the fact `Counter` cannot be decreased.
My intention was to avoid https://github.com/jakub-dzon/k4e-operator/pull/46/files#diff-0e245eaea02ac9a12c64a8b1e4ec99514a84e7066ca56aac23f9f0591c71fab4R69 for listing and summerizing the amount of edge-devices in each loop.

Hence instead of using the `createdEdgeDevices.Set(float64(value))` to use `createdEdgeDevices.Inc())` when registering new device (or when creating if it is implemented in the edge-device controller) and to use `createdEdgeDevices.Decc())` when edge device is de-registered.

controllers/edgedevice_controller.go
```diff
@@ -73,6 +75,13 @@ func (r *EdgeDeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{Requeue: true}, err
 	}
 	logger.Info("Reconciling", "edgeDevice", edgeDevice)
+	numberOfEdgeDevices, err := r.EdgeDeviceRepository.NumberOfEdgeDevices(ctx)
+	if err != nil {
+		logger.Error(err, "Cannot get number of edge devices")
+	} else {
+		logger.Info("******************* @ Updating number of edge devices *********")
```
jakub-dzon, line 82: If the code is supposed to be merged, please drop the stars.

jakub-dzon (changes requested):

internal/metrics/metrics.go
masayag: Specifically for the scale testing I'd like to capture the number of successful registration vs failures.
For that purpose counter can be a good option (the numbers are expected only to grow).

controllers/edgedeployment_controller.go
```diff
@@ -65,6 +67,13 @@ type ConcurrentFunc func([]managementv1alpha1.EdgeDevice) []error
 func (r *EdgeDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 	logger.Info("Reconciling", "request", req)
+	numberOfEdgeDeployments, err := r.EdgeDeploymentRepository.NumberOfEdgeDeployments(ctx)
+	if err != nil {
+		logger.Error(err, "Cannot get number of edge devices")
+	} else {
+		logger.Info("******************* @ Updating number of edge devices *********")
```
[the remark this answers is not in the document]
bardielle: Sure

internal/metrics/metrics.go
bardielle: After discussing about it we decided to use Counter for registration (succeeded and failures), and OBCs creation 
The creation of edge devices and edge deployment not relevant for the scaling tests

internal/repository/edgedeployment/edgedeployment.go
```diff
@@ -44,6 +45,16 @@ func (r *CRRespository) ListForEdgeDevice(ctx context.Context, name string, name
 	return edl.Items, nil
 }
 
+func (r *CRRespository) NumberOfEdgeDeployments(ctx context.Context) (int, error) {
+	edl := v1alpha1.EdgeDeploymentList{}
+
+	err := r.client.List(ctx, &edl)
```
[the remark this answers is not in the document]
bardielle: You're right
I removed it cause we decided that for investigating the scailing issue we will use registration information only

internal/repository/edgedeployment/edgedeployment.go
```diff
@@ -13,6 +13,7 @@ import (
 //go:generate mockgen -package=edgedeployment -destination=mock_edgedeployment.go . Repository
 type Repository interface {
 	ListForEdgeDevice(ctx context.Context, name string, namespace string) ([]v1alpha1.EdgeDeployment, error)
+	NumberOfEdgeDeployments(ctx context.Context) (int, error)
```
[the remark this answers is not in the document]
bardielle: I changed the functions names (Inc+the metric name)

config/manager/manager.yaml
bardielle: That's what I saw in the documentation and examples

config/manager/manager.yaml
masayag: iiuc, kustomize is responsible to replace it with the actual namespace in the created resource: https://github.com/jakub-dzon/k4e-operator/blob/main/config/default/kustomization.yaml#L2

internal/yggdrasil/yggdrasil.go
```diff
@@ -263,6 +267,7 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 			return operations.NewPostDataMessageForDeviceInternalServerError()
```
masayag, line 267: Need to add `h.metrics.IncFailedRegistrationEdgeDevice()` here as well

internal/metrics/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+	IncCreatedOBCs()
+}
+
+func New() Metrics {
+	return &metricsImpl{}
+}
+
+type metricsImpl struct{}
+
+func (m *metricsImpl) IncRegisteredEdgeDevice() {
```
jakub-dzon, line 60: Is this metric supposed to represent just number of registrations or current number of edgedevices under operator's purview?

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,69 @@
...
+		m.IncRegisteredEdgeDevice()
+	}
+	for  i := 0; i < numberOfFailedToRegisterValue; i++ {
+		m.IncFailedRegistrationEdgeDevice()
+	}
+	for  i := 0; i < numberOfCreatedOBCsValue; i++ {
+		m.IncCreatedOBCs()
+	}
+
+	It("correctly passes calls to the collectors", func() {
```
jakub-dzon, line 45: Please use 
```
// given
// when
// then 
```
template for tests to highlight their different aspects (input, execution, verification)

jakub-dzon (changes requested):

internal/metrics/metrics.go
bardielle: Just number of registrations

jakub-dzon: In that case the name of the method name is a little confusing to me - it can mean both. WDYT about making the name a bit more explicit; like `IncEdgeDeviceRegisteredSuccessfully` and `k4e_operator_edge_devices_registered_successfully` or `IncEdgeDeviceSuccessfulRegistration` and `k4e_operator_edge_devices_successful_registration` or simply aligning it with the metric name: `IncEdgeDevicesWereRegistered`.

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,73 @@
...
+	return nil
+}
+
+var _ = Describe("Metrics", func() {
+	m := New()
+
+	It("correctly passes calls to the collectors", func() {
+		//when
+		for  i := 0; i < numberOfRegisteredValue; i++ {
+			m.IncRegisteredEdgeDevice()
```
jakub-dzon, line 39: Here three different methods are tested in one test cases; they should have their own, separate ones. The code as it is invites future developers adding more metrics to simply add their tests  here and after a while this one test might become a hundreds-of-lines monster. 

The verification part is also not so easy to understand at first glance.

As per [contributing guidelines](https://github.com/jakub-dzon/k4e-operator/blob/main/CONTRIBUTING.md#testing) any logic and conditionals in tests should be avoided.

I'm sorry for not pointing this out in my previous comment.

bardielle: I already tried to split it into 3 different test but I saw several issues - first the metrics object in the test cannot be initialized between the tests and also the return value from that line `data, err := metrics.Registry.Gather()` is a struct of all of the metrics so need to go over each one of them anyway

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,71 @@
+package metrics
```
jakub-dzon, line 1: Please change to `metrics_test`

internal/metrics/metrics_test.go
jakub-dzon: Inability to have fresh metrics registry between tests is unfortunate, but it's not a big problem if the tests don't depend on each other.

This is test code and we should prioritize readability over performance/saving a few loops. The verification code: https://github.com/jakub-dzon/k4e-operator/pull/46/files#diff-8ea0d64df378dd743e9e7106592591a8568f325916af29bf4980054b6373b370R49 and https://github.com/jakub-dzon/k4e-operator/pull/46/files#diff-8ea0d64df378dd743e9e7106592591a8568f325916af29bf4980054b6373b370R62 is a bit cryptic.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// When adding metric names, see https://prometheus.io/docs/practices/naming/#metric-names
+const (
+	edgeDeviceSuccessfulRegistration = "k4e_operator_edge_devices_successful_registration"
+	edgeDeviceFailedRegistration     = "k4e_operator_edge_devices_failed_registration"
+	createdOBCQuery                  = "k4e_operator_created_obc"
```
pkliczewski, line 12: What do you think about adding metrics related to `edgeDeployment`?

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,79 @@
...
+	Expect(m).ToNot(BeNil(), "metric for %s could not be found", query)
+	Expect(m.Metric).To(HaveLen(1))
+	Expect(m.Metric[0].Counter).ToNot(BeNil())
+	Expect(m.Metric[0].Counter.Value).ToNot(BeNil())
+	Expect(*m.Metric[0].Counter.Value).To(BeEquivalentTo(value))
+}
+var _ = Describe("Metrics", func() {
+	m := metrics.New()
+
+	Context("IncEdgeDeviceSuccessfulRegistration", func() {
```
jakub-dzon, line 48: I think that and other `Context` demarcations are not needed; the names of the metrics are already in `It`. What might be useful is establishing context for EdgeDevice metrics and put all 3 test cases in it.

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,79 @@
...
+		It("correctly passes calls to the IncEdgeDeviceSuccessfulRegistration", func() {
```
jakub-dzon, line 49: I really like the look of these tests - short and clear. 
Nice refactoring!

bardielle: thanks

internal/metrics/metrics.go
bardielle: I added and removed it because @masayag and I weren't sure how it can help us for the test scaling... 
@masayag

masayag: I miss to see how the notation of `edgedeployment` on itself is useful for us.
Theoretically, we can have tons of edgedeployments, without an edge device to be played on.
Therefore, when detached from the context of edge-device, the edge-deployments won't provide useful information.

However, they are meaningful in the context of edge-device. We can monitor how many edge-deployments are successfully running (or failed to run), to create an alert based on specific rule (e.g. if failed to run a specific edgedeployment on % of the devices might indicate on an issue with the deployment itself).
however, the latter requires a design to define the requirements for it (for observability purpose).

config/manager/manager.yaml
```diff
@@ -62,6 +62,26 @@ spec:
       protocol: TCP
       port: 8888
       targetPort: 8888
+    - name: metrics
+      protocol: TCP
+      port: 8080
```
masayag, line 67: @eloycoto is your work affected by this change ?

jakub-dzon (approved):

config/manager/manager.yaml
eloycoto: No

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	m := findMetric(data, query)
+	Expect(m).ToNot(BeNil(), "metric for %s could not be found", query)
+	Expect(m.Metric).To(HaveLen(1))
+	Expect(m.Metric[0].Counter).ToNot(BeNil())
+	Expect(m.Metric[0].Counter.Value).ToNot(BeNil())
+	Expect(*m.Metric[0].Counter.Value).To(BeEquivalentTo(value))
+}
+var _ = Describe("Metrics", func() {
+	m := metrics.New()
```
eloycoto, line 46: Just a thing, I will move this to the BeforeEach section. Mainly because you want to have a clear object on Each Test and to avoid future test flakes.

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func validateMetric(query string, value int){
+	data, err := ctrlmetrics.Registry.Gather()
+	Expect(err).NotTo(HaveOccurred())
+
+	m := findMetric(data, query)
+	Expect(m).ToNot(BeNil(), "metric for %s could not be found", query)
+	Expect(m.Metric).To(HaveLen(1))
+	Expect(m.Metric[0].Counter).ToNot(BeNil())
```
eloycoto, line 41: Just a nitpick, if you use NotTo in the line 35, please use the same on these lines. (It's ok to keep it)

internal/metrics/metrics_test.go
bardielle: Unfortunately metrics object cannot be cleared

masayag (approved):

PR #47, opened 2021-12-08
support secrets

ydayagi:
parse podSpecs, read the referenced secrets, send to device

https://issues.redhat.com/browse/ECOPROJECT-221


jakub-dzon (changes requested):
Nice work! I like that the added test cases cover different scenarios.

Just a few code style-related comments.

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +429,114 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+	// create map of secret names and keys
+	secretMap := secretMapType{}
+	for _, deployment := range deployments {
+		podSpec := deployment.Spec.Pod.Spec
+		allContainers := append(podSpec.InitContainers, podSpec.Containers...)
+		for _, container := range allContainers {
+			extractSecretsFromContainer(&container, secretMap)
+		}
+	}
+	fmt.Println(secretMap)
```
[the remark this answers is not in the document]
ydayagi: oops. I'm getting careless. I reviewed my code but missed it. This is why we have code review :-)

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +429,114 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+		list = append(list, &models.Secret{
+			Data: string(dataJson),
+			Name: secretObj.Name,
+		})
+	}
+
+	return list, nil
+}
+
+type keyMapType = map[string]interface{}
```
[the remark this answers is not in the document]
ydayagi: Thanks. I didn't know what u guys will prefer, top or next to function, so I just put it there so u all can comment. will change

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +429,114 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+					continue
+				}
+			}
+			return nil, fmt.Errorf("missing secret key. secret: %s. key: %s", name, key)
+		}
+		secretObjs = append(secretObjs, secretObj)
+	}
+
+	// add secrets to list
+	for _, secretObj := range secretObjs {
```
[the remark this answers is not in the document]
ydayagi: I can try, but I think it is easier to understand when separating to different actions/steps. If I make it a single loop u get all the logic in a single iteration. I think it makes it less readable. I'll try with functions.

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +429,114 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+			}
+			return nil, err
+		}
+		for key := range keys {
+			if secretObj.Data != nil {
+				if _, ok := secretObj.Data[key]; ok {
+					continue
+				}
+			}
+			return nil, fmt.Errorf("missing secret key. secret: %s. key: %s", name, key)
```
pkliczewski, line 465: I wonder why we stop iterating over the secrets? What do you think about `warn` with descriptive message and continue iterating?

pkliczewski (commented):
How do we watch for secret changes?

internal/yggdrasil/yggdrasil.go
ydayagi: I followed the same method as with workloads/deployments. I do agree that there is room to re-think the way we handle errors.

ydayagi: also...if we do not stop here then it will fail in the device. Since EdgeDeployment is used by many devices it means that many devices will have failures.

ydayagi:
> 

I did not handle that yet. We have a discussion about it in Jira. I think we need to decide whether we are sure we want to watch secrets and whether to do it as part of this story or not.

jakub-dzon:
> > 
> 
> I did not handle that yet. We have a discussion about it in Jira. I think we need to decide whether we are sure we want to watch secrets and whether to do it as part of this story or not.

Secret watching is not needed - the secrets are retrieved in their most current state when device asks for its configuration and knowing in the operator that some secret changed in between does not give us anything. We just need to make sure on the device side that changes to the secrets are applied.

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +432,134 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
 
 	return err
 }
+
+func (h *Handler) readAndValidateSecret(ctx context.Context, secretName, secretNamespace string, secretKeys keyMapType) (*corev1.Secret, error) {
```
machacekondra, line 436: Can we have those secret manipulation functions moved to some `secrets.go` module?

ydayagi: these are specific to yggdrasil.go. It's just for making the functions smaller.
Let's wait till we see how we deal with other resources and then we can have a 'pod spec parsing' package

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +432,134 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+}
+
+// return value maps a secret name to its mandatory keys. nil keys indicates optional secret
+func extractSecretsFromContainer(container *corev1.Container, secretMap secretMapType) {
+	extractSecretsFromEnvFrom(container.EnvFrom, secretMap)
+	extractSecretsFromEnv(container.Env, secretMap)
+}
+
+func extractSecretsFromEnvFrom(envFrom []corev1.EnvFromSource, secretMap secretMapType) {
+	if len(envFrom) == 0 {
```
eloycoto, line 512: why this len? No need at all there is nothing to range no?

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +432,134 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+		return
+	}
+	for _, envFrom := range envFrom {
+		if envFrom.SecretRef != nil {
```
eloycoto, line 516: what about `envFRom.Secretref == nil {continue}` so things are a bit flatten?

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +432,134 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+				} else {
+					secretMap[envFrom.SecretRef.Name] = keyMapType{}
+				}
+			}
+		}
+	}
+}
+
+func extractSecretsFromEnv(env []corev1.EnvVar, secretMap secretMapType) {
+	if len(env) == 0 {
```
eloycoto, line 537: this len is no needed at all no?

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +432,134 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+	for _, envVar := range env {
+		if envVar.ValueFrom != nil && envVar.ValueFrom.SecretKeyRef != nil {
+			keyRef := envVar.ValueFrom.SecretKeyRef
+			var optional bool
+			if keyRef.Optional != nil {
+				optional = *keyRef.Optional
+			}
+			if keys, ok := secretMap[keyRef.Name]; ok {
+				if optional {
+					break
```
eloycoto, line 549: this is a break, or a continue?

internal/yggdrasil/yggdrasil_test.go
```diff
@@ -586,6 +592,643 @@ var _ = Describe("Yggdrasil", func() {
...
+			device.Status.Deployments = []v1alpha1.Deployment{{Name: "workload1"}}
+
+			edgeDeviceRepoMock.EXPECT().
+				Read(gomock.Any(), deviceName, testNamespace).
+				Return(device, nil).
+				Times(1)
+
+			secretName := "test"
+			secretNamespacedName := types.NamespacedName{Namespace: device.Namespace, Name: secretName}
+			podData := v1alpha1.Pod{
```
eloycoto, line 668: This podData can be a gunction like `getDevice` no?

internal/yggdrasil/yggdrasil_test.go
```diff
@@ -586,6 +592,643 @@ var _ = Describe("Yggdrasil", func() {
...
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			}
+
+			deploymentData := &v1alpha1.EdgeDeployment{
```
eloycoto, line 688: The same as podData, can be an external function like getDevice()

internal/yggdrasil/yggdrasil_test.go
```diff
@@ -586,6 +592,643 @@ var _ = Describe("Yggdrasil", func() {
...
+							Env: []corev1.EnvVar{
+								{
+									Name: "test",
+									ValueFrom: &corev1.EnvVarSource{
+										SecretKeyRef: &corev1.SecretKeySelector{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: secretName,
+											},
+											Key:      "key",
+											Optional: &booltrue,
```
eloycoto, line 753: why you defined boolTrue and not using true here?

internal/yggdrasil/yggdrasil.go
eloycoto: I would create another package for it too.

internal/yggdrasil/yggdrasil_test.go
```diff
@@ -586,6 +592,643 @@ var _ = Describe("Yggdrasil", func() {
...
+			device.Status.Deployments = []v1alpha1.Deployment{{Name: "workload1"}}
+
+			edgeDeviceRepoMock.EXPECT().
+				Read(gomock.Any(), deviceName, testNamespace).
+				Return(device, nil).
+				Times(1)
+
+			secretName := "test"
+			secretKey := "key"
+			secretNamespacedName := types.NamespacedName{Namespace: device.Namespace, Name: secretName}
```
eloycoto, line 803: All those variables are almost shared within all the test, I would move this declaration to the beforeEach function.

eloycoto (changes requested):
Test IMHO need a bit of refactoring to make it a bit simpler. Like getPod, get Deployment

internal/yggdrasil/yggdrasil.go
ydayagi: right

internal/yggdrasil/yggdrasil.go
ydayagi: right

internal/yggdrasil/yggdrasil.go
ydayagi: ok

internal/yggdrasil/yggdrasil.go
ydayagi: great catch. thanks.

internal/yggdrasil/yggdrasil_test.go
ydayagi: no because I use differnet podData objects and they are very different

internal/yggdrasil/yggdrasil_test.go
ydayagi: It requires a pointer. Can use constant as pointer

internal/yggdrasil/yggdrasil_test.go
ydayagi: not all tests. only negative tests and even there, I think. BeforeEach and such should be used fo meaningful test content and not just a few constant vars. If it was up to me I wouldn't use vars. I would just use constants directly where necessary.

internal/yggdrasil/yggdrasil_test.go
ydayagi: all tests set deploymentData within the test. this is because, unlike divice object, this is the heart of the tests. Not good to take it out.

internal/yggdrasil/yggdrasil_test.go
eloycoto: Ok, then defined on the Context of the test, no?

internal/yggdrasil/yggdrasil_test.go
eloycoto: is all similar, I'm trying to make it simpler to read and not duplicating code. As other parts of the test.

internal/yggdrasil/yggdrasil_test.go
```diff
@@ -586,6 +592,643 @@ var _ = Describe("Yggdrasil", func() {
 			Expect(eventsRecorder.Events).To(Receive(ContainSubstring("Auth file secret")))
 		})
 
+		It("Secrets reading failed", func() {
+			// given
+			deviceName := "foo"
```
eloycoto, line 597: This deviceName&device can be defined on the `beforeaEach` section.

internal/yggdrasil/yggdrasil_test.go
eloycoto: Then you can define in a `context("Negative test", func())` and simplify the tests.

internal/yggdrasil/yggdrasil.go
```diff
@@ -416,3 +432,129 @@ func (h *Handler) setStorageConfiguration(ctx context.Context,
...
+	if err != nil {
+		if errors.IsNotFound(err) && optional {
+			return nil, nil
+		}
+		return nil, err
+	}
+	for key := range secretKeys {
+		if secretObj.Data != nil {
+			if _, ok := secretObj.Data[key]; ok {
+				continue
```
eloycoto, line 449: This part it's not tested, no?

eloycoto (changes requested):

internal/yggdrasil/yggdrasil_test.go
ydayagi: We cna define it anywhere. It doesn't matter where and whether once or multiple times.

internal/yggdrasil/yggdrasil_test.go
ydayagi: There's no need

internal/yggdrasil/yggdrasil.go
ydayagi: We don't test all use-cases/permutations in yggdrasil.go and in the entire project in general

internal/yggdrasil/yggdrasil_test.go
ydayagi: not every line of code needs to be avoided.

jakub-dzon (approved):

internal/yggdrasil/yggdrasil.go
pkliczewski: Good catch @ydayagi! We can improve this. It is  good practice to find bugs with unit tests rather than production.

pkliczewski: Let's improve coverage. Would you like to do it in this PR or new one?

ydayagi: this PR

internal/yggdrasil/yggdrasil.go
ydayagi: @machacekondra when u get to configMaps, if you see fit to put all these functions in a single package then ping me and I'll do it for you.

internal/yggdrasil/yggdrasil.go
ydayagi: this code is already covered. possitive test is "Secrets reading succeeded". Negative test is "Secrets missing secret key". So we verify that if key is missing continue does not happen. we verify that if key is present there's no error.

pkliczewski: I think new PR would be better to improve test coverage but it is up to you, This PR is OK as well.

ydayagi: > This part it's not tested, no?

I was mistaken. It is covered

eloycoto (approved):

pkliczewski (approved):

pkliczewski:
@masayag @machacekondra any other comments or are we OK to merge?

machacekondra (approved):

PR #48, opened 2021-12-08
Reduce verbosity detailed messages

masayag:

### DIFF
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -184,7 +184,7 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 		if err != nil {
 			return operations.NewPostDataMessageForDeviceBadRequest()
 		}
-		logger.Info("received heartbeat", "content", heartbeat)
+		logger.V(1).Info("received heartbeat", "content", heartbeat)
 		edgeDevice, err := h.deviceRepository.Read(ctx, params.DeviceID, h.initialNamespace)
 		if err != nil {
 			if errors.IsNotFound(err) {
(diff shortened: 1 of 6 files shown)


internal/yggdrasil/yggdrasil.go
```diff
@@ -184,7 +184,7 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 		if err != nil {
 			return operations.NewPostDataMessageForDeviceBadRequest()
 		}
-		logger.Info("received heartbeat", "content", heartbeat)
+		logger.V(1).Info("received heartbeat", "content", heartbeat)
```
jakub-dzon, line 187: Why the increase of verbosity here?

internal/yggdrasil/yggdrasil.go
```diff
@@ -236,7 +236,7 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 		if err != nil {
 			return operations.NewPostDataMessageForDeviceBadRequest()
 		}
-		logger.Info("received registration info", "content", registrationInfo)
+		logger.V(1).Info("received registration info", "content", registrationInfo)
```
jakub-dzon, line 239: And here?

internal/yggdrasil/yggdrasil.go
masayag: Based on https://github.com/jakub-dzon/k4e-operator#implementation-details, it decreases the verbosity
and from the documentation:
```
	// V returns an Logger value for a specific verbosity level, relative to
	// this Logger.  In other words, V values are additive.  V higher verbosity
	// level means a log message is less important.  It's illegal to pass a log
	// level less than zero.
	V(level int) Logger
```
Hence, we shouldn't see the details for those messages when the log level is set to `info`

jakub-dzon: Ok, my bad; I confused zap logging priority with logr verbosity level.

jakub-dzon (approved):

pkliczewski (approved):

PR #51, opened 2021-12-10
Add context to each log message of setupLog

gciavarrini:

### DIFF
--- a/main.go
+++ b/main.go
@@ -126,7 +126,7 @@ func init() {
 
 func main() {
 	err := envconfig.Process("", &Config)
-	setupLog = ctrl.Log.WithName("setup")
+	setupLog = ctrl.LoggerFrom(context.TODO())
 
 	if err != nil {
 		setupLog.Error(err, "unable to process configuration values")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -145,10 +145,16 @@ func main() {
...
+	logger := zap.New(zap.UseFlagOptions(&opts)).WithName("setup")
 	ctrl.SetLogger(logger)
 
-	setupLog = ctrl.Log
+	operatorNamespace, err := getOperatorNamespace()
+	if err != nil {
+		setupLog.Error(err, "cannot get the operator namespace")
+		os.Exit(1)
+	}
+	setupLog = ctrl.Log.WithValues("namespace", operatorNamespace)
```
masayag, line 156: I don't think we should record the namespace in each log message. This namespace suppose to be same through the entire run of the operator and should be known to the admin before viewing the log file (`kubectl logs k4e-podname -n k4e-namespace`)

gciavarrini: You're right! Is there any other contextual information you think is worth adding?

masayag: not for the setupLog

PR #54, opened 2021-12-13
Namespace issues

eloycoto:
Hi, 

Two commits here:

1) main.go move operatorNamespace variable set on init phase. So always use the global var. @gciavarrini maybe I break something here. 
2) Makefile: add ns creation on make run to not fail on MTLS. 

Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,12 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Default k4e-Operator  namespace
+K4E_OPERATOR_NAMESPACE ?= "k4e-operator-system"
+
+# Set quiet mode by default
+Q=@
+
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # This is a requirement for 'setup-envtest.sh' in the test target.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
@@ -124,6 +130,7 @@ build: generate fmt vet ## Build manager binary.
 	go build -mod=vendor -o bin/manager main.go
 
 run: manifests generate fmt vet ## Run a controller from your host.
+	$(Q) kubectl create ns $(K4E_OPERATOR_NAMESPACE) 2> /dev/null || exit 0
 	OBC_AUTO_CREATE=false LOG_LEVEL=debug go run -mod=vendor ./main.go
 
 docker-build: test ## Build docker image with the manager.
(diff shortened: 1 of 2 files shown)


pkliczewski (approved):

main.go
```diff
@@ -117,6 +118,14 @@ func init() {
 	utilruntime.Must(managementv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(obv1.AddToScheme(scheme))
 	utilruntime.Must(routev1.AddToScheme(scheme))
+
+	ns, err := getOperatorNamespace()
+	if err == nil {
+		operatorNamespace = ns
+	} else {
+		operatorNamespace = defaultOperatorNamespace
```
gciavarrini, line 126: The `getOperatorNamespace` function returns the `defaultOperatorNamespace` in case of "_In cluster_" run.
In case of "_Out of the cluster_" run, if the namespace has not been created and we use the `defaultOperatorNamespace` then the get `ConfigMaps` will fail. If it is not possible to find a namespace in the case of _Out of the cluster_ run then the operator should fail.

Makefile
```diff
@@ -47,6 +47,12 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Default k4e-Operator  namespace
+K4E_OPERATOR_NAMESPACE ?= "k4e-operator-system"
```
bardielle, line 51: There are a lot of places in the makefile that use the explicit name of the namespace 
maybe a good refactoring is to replace all of them... 
Also, there is assumption in our code that the default namespace called "k4e-operator-system" (for example main.go and https://github.com/jakub-dzon/k4e-operator/blob/main/config/default/kustomization.yaml#L2)
There is a way to reuse that parameter?

eloycoto: For the main.go I do not think so. This way only uses one variable instead of multiple. 
For the makefile, I cannot find where you're referring to. The idea to have in a variable is to be able to do `make run  K4E_OPERATOR_NAMESPACE=newNS`
For the kustamization, that it's a variable AFAIK, no?

bardielle: You're right I looked into the README file, my mistake

jakub-dzon (approved):

main.go
eloycoto: Check now, I think that I fixed :-)

Makefile
eloycoto: :+1:

gciavarrini (approved):
LGTM

PR #60, opened 2021-12-17
🧪 Integration test POC 🚀

eloycoto:
This commit introduced a POC CI testing framework; this is not a
complete solution; it is a POC to show if we want to follow this way.

The main goal here is to use the kind cluster that GH actions are
running and use the Bats-core testing framework[0] to create edge devices
using an HTTP client(curl). Using yggradsil and device-worker will be
hard due to the nature of the workers, but we can try.

Each test file can contain a small test; the idea is to have a
lightweight CI on each PR, like:

- Creating an edgedevice and checking that all new MTLS workflow will
  work.
- Trigger a small Jmetter scenario that we can reuse, but with fewer
  clients.
- Run operator with race conditions and deadlocks checks
- Testing a deployment is correctly triggered, etc..
- Be able to test getting starting guides.

I know it's bash code, but it is easy to understand and keep up to date.
(I hope I do not regret about this)

[0] https://bats-core.readthedocs.io/en/stable/tutorial.html#your-first-test

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -10,6 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
+        with:
+          submodules: recursive
       - uses: engineerd/setup-kind@v0.5.0
       - name: Testing on KinD
         run: |
@@ -23,3 +25,7 @@ jobs:
           if [ $retVal -ne 124 ]; then
             echo "Make run failed"; exit 1
           fi
+
+      - name: Integration tests
+        run: |
+          timeout 10m ./tests/bats/bin/bats --verbose-run tests/registration.bats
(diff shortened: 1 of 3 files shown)


jakub-dzon:
I'm a bit afraid of using bash for testing; been there with CNV Chaos testing (not with bats, though) and it's been pain to debug. WDYT about implementing it in Go - CNV uses go for upstream e2e tests (and we used it in the [vm-import-operator](https://github.com/kubevirt/vm-import-operator/tree/master/tests/ovirt)).

masayag:
This is a great idea and I like it.
I also think that it would be safer to test the code using golang vs bash. It will allow us using the API and generated code based on our swagger file and avoid mismatch between the API and the tests.

pkliczewski:
I agree with @jakub-dzon and @masayag let's use go instead of bash.

eloycoto:
Closing this then!

PR #61, opened 2021-12-21
CI: Add lint check on Github Actions

eloycoto:
This commit adds a new check on the Github Action, where a lint is
passed on all code to make sure that there  is no mistakes.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -7,7 +7,6 @@ on:
     branches: [ main ]
 
 jobs:
-
   build:
     runs-on: ubuntu-latest
     steps:
@@ -32,3 +31,14 @@ jobs:
       with:
         name: test-coverage
         path: coverage.html
+
+
+  lint-code:
+    name: Lint code
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v2.5.1
--- a/internal/images/registry-auth.go
+++ b/internal/images/registry-auth.go
@@ -2,8 +2,8 @@ package images
 
 import (
 	"context"
-	"errors"
 	"fmt"
+
 	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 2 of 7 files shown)


jakub-dzon (approved):
Tightening the screws, aren't we?:D Very good!

eloycoto:
yeah!

.github/workflows/go.yml
```diff
@@ -32,3 +31,14 @@ jobs:
       with:
         name: test-coverage
         path: coverage.html
+
+
+  lint-code:
+    name: Lint code
+    runs-on: ubuntu-latest
```
ydayagi, line 38: I don't know how to add a github action. So my questions are for learning. What is the reason for specifying ubuntu-latest?

internal/images/registry-auth.go
```diff
@@ -29,7 +29,7 @@ func (r *RegistryAuth) GetAuthFileFromSecret(ctx context.Context, namespace, nam
 	}
 	authFile, found := secret.Data[".dockerconfigjson"]
 	if !found {
-		return "", errors.New(fmt.Sprintf(".dockerconfigjson not found in %s/%s Secret", namespace, name))
+		return "", fmt.Errorf(".dockerconfigjson not found in %s/%s Secret", namespace, name)
```
ydayagi, line 32: is this something that golint brought up? cause i also used 'errors.New' here and there.

.github/workflows/go.yml
eloycoto: it's the flavour that GH uses.

internal/images/registry-auth.go
eloycoto: yes, Errors.New without fmt.Sprintf() is ok.

main.go
```diff
@@ -284,7 +288,9 @@ func main() {
 		}
 
 		//@TODO This is a hack to keep compatibility now, to be deleted.
-		go http.ListenAndServe(fmt.Sprintf(":%v", Config.HttpPort), h)
+		go func() {
```
ydayagi, line 291: just for me to understand/learn. what is the reason to switch from single line to function block?

.github/workflows/go.yml
ydayagi: great, thanks.

internal/images/registry-auth.go
ydayagi: ahhhhh it's the combination of the two that it doesn't like. ok. go figure.
Thanks again. good to know.

main.go
eloycoto: err was not checked, and lint complains about. 

I added this mainly because it's going to be deleted in the following days.

ydayagi: intersting. thanks.

ydayagi (approved):

PR #62, opened 2021-12-21
Metrics: Added metrics configuration

eloycoto:
This commits adds the new metrics information on the spec. The value
will be `port` and `path`.

List of changes:

- Changed CRD definition.
- Update workload API response definition.
- If port is not defined, no metrics will be used.

Related to: ECOPROJECT-425


api/v1alpha1/edgedeployment_types.go
```diff
@@ -32,12 +32,18 @@ type EdgeDeploymentSpec struct {
...
 	Data            *DataConfiguration            `json:"data,omitempty"`
 	ImageRegistries *ImageRegistriesConfiguration `json:"imageRegistries,omitempty"`
+	Metrics         *PrometheusMetrics            `json:"metrics,omitempty"`
 }
 
 type ImageRegistriesConfiguration struct {
 	AuthFileSecret *ObjectRef `json:"secretRef,omitempty"`
 }
 
+type PrometheusMetrics struct {
```
jakub-dzon, line 42: Let's keep it technology-agnostic: `MetricsConfiguration`

internal/yggdrasil/yggdrasil.go
```diff
@@ -405,6 +405,13 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
 				AuthFile: authFile,
 			}
 		}
+
+		if spec.Metrics != nil && spec.Metrics.Port > 0 {
+			workload.Metrics = &models.Metrics{
+				Path: spec.Metrics.Path,
```
jakub-dzon, line 411: What if `Port>0` and `Path==""`?

jakub-dzon (changes requested):

internal/yggdrasil/yggdrasil.go
eloycoto: I guess that path empty is /, no?

eloycoto:
@jakub-dzon  and all, what about if one Deployment defines two pods? will require the two containers to listen at the same port/path?

Regards

internal/yggdrasil/yggdrasil.go
jakub-dzon: :+1:

jakub-dzon:
Good point. We don't support deployments and might not be implementing that support at all, but we might have similar problem already - pod can have multiple containers. What if there is a need to scrape metrics from more than one? What if we need to scrape from only one of many? IMO we can enforce same path/port for all as the simplest solution, but selective metric scraping might be a problem.

jakub-dzon:
We might also need to allow users to define scraping interval to allow for different metrics resolutions.

eloycoto:
Added the changes + Kubebuilder annotations for min/max.

internal/1.txt
```diff
@@ -0,0 +1 @@
+11
```
jakub-dzon, line 1: That file has probably been added by mistake.

internal/images/gomock_reflect_782103550/prog.go
```diff
@@ -0,0 +1,64 @@
+package main
```
jakub-dzon, line 1: That file has probably been added by mistake.

jakub-dzon (changes requested):
There are some stray temporary files that need to be removed.

jakub-dzon (approved):
LGTM. 
Please rebase.

pkliczewski (approved):

swagger.yaml
```diff
@@ -267,6 +270,36 @@ definitions:
...
         type: string
 
+  metrics:
+    type: object
+    properties:
+      path:
+        description: Path to use when retrieving metrics
+        type: string
+      port:
+        type: integer
```
masayag, line 280: WDYT of including also `format: int32` which is suitable for the port range?

api/v1alpha1/edgedeployment_types.go
```diff
@@ -26,18 +26,50 @@ import (
...
 
+type MetricsConfigEntity struct {
+	// Path to use when retrieving metrics
+	// +kubebuilder:default=/
+	Path string `json:"path,omitempty"`
+
+	// Port to use when retrieve the metrics
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	Port int `json:"port,omitempty"`
```
masayag, line 50: int32 should fit this purpose.

api/v1alpha1/edgedeployment_types.go
```diff
@@ -26,18 +26,50 @@ import (
...
+
+type ContainerMetricsConfiguration struct {
+	// Path to use when retrieving metrics
+	// +kubebuilder:default=/
+	Path string `json:"path,omitempty"`
+
+	// Port to use when retrieve the metrics
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=65535
+	Port int `json:"port,omitempty"`
```
masayag, line 63: int32 should fit this purpose.

internal/yggdrasil/yggdrasil.go
```diff
@@ -422,6 +422,26 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
...
+			for container, metricConf := range spec.Metrics.Containers {
+				containers[container] = models.ContainerMetrics{
+					Disabled: metricConf.Disabled,
+					Port:     int64(metricConf.Port),
+					Path:     metricConf.Path,
+				}
+				addedContainers = true
+			}
+			if addedContainers {
+				workload.Metrics.Containers = containers
```
masayag, line 442: on the worker side workload.Metrics.Containers with `nil` or zero value behaves differently?

eloycoto: no, will range over Metrics.Containers.

swagger.yaml
eloycoto: Done! Sorry!

masayag (approved):

PR #63, opened 2021-12-22
Expose Operator's HTTP endpoint externally

arielireni:
Create Ingress resource on make deploy, which will make HTTP endpoint accessible externally out-of-the-box and port-forwarding is not required.

On OpenShift cluster:
1. Deploy with `IMG=<image registry and tag> TARGET=ocp make deploy`.
2. Get HTTP server address by running:
`HTTP_SERVER=$(oc get routes k4e-operator-controller-manager -n k4e-operator-system --no-headers -o=custom-columns=HOST:.spec.host)`.
3. Start yggdrasil with host/port value from step 2: `sudo go run ./cmd/yggd --log-level info --transport http --cert-file /etc/pki/consumer/cert.pem --key-file /etc/pki/consumer/key.pem --client-id-source machine-id --http-server $HTTP_SERVER`.

On Minikube cluster:
1. Deploy with `IMG=<image registry and tag> HOST=<host name> TARGET=k8s make deploy`.
2. Add to `/etc/hosts`: `<minikube ip> <host name>`.
3. Start yggdrasil with from the yggdrasil repository directory: `sudo go run ./cmd/yggd --log-level info --transport http --cert-file /etc/pki/consumer/cert.pem --key-file /etc/pki/consumer/key.pem --client-id-source machine-id --http-server <host name>`.

If `HOST` isn't specified, the default value is `k4e-operator.srv`.
If `TARGET` isn't specified, the default value is `k8s`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,10 @@ BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+# Cluster type - k8s/ocp
+TARGET ?= k8s
+# Host name for ingress creation
+HOST ?= k4e-operator.srv
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -80,7 +84,7 @@ help: ## Display this help.
 ##@ Development
 
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	@$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 generate-tools:
 ifeq (, $(shell which mockery))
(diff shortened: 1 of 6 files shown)


config/manager/manager.yaml
```diff
@@ -89,3 +89,23 @@ spec:
...
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: operator-ingress
+  namespace: k4e-operator-system
+  annotations:
+    nginx.ingress.kubernetes.io/rewrite-target: /
+spec:
+  rules:
+    - host: operator.info
```
eloycoto, line 102: This is correct, or should be a parameter somewhere?

masayag: I'd exclude this section from this file, else its output will be used to generate the operator yaml which isn't desired for all cases.
As Eloy suggested, for dev-env, user can provide the hostname as parameter to be used or fallback to a default

masayag:
Regarding the tradeoff from the PR message, for OCP we could replace the ingress CR with Route by invoking:
```
oc expose svc k4e-operator-controller-manager -n k4e-operator-system
```

masayag (changes requested):

config/manager/manager.yaml
arielireni: done

config/networking/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+resources:
```
jakub-dzon, line 1: Neither ingress nor route will be present in the release yaml and I think one of them (probably ingress, as its universal) should be there.

masayag (changes requested):

Makefile
```diff
@@ -147,17 +147,39 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+TARGET ?= k8s
+HOST ?= operator.info
```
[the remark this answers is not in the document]
arielireni: done

Makefile
```diff
@@ -147,17 +147,39 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+TARGET ?= k8s
```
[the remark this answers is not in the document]
arielireni: done

Makefile
```diff
@@ -147,12 +147,39 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
...
+	cd config/networking && $(KUSTOMIZE) edit add resource ingress.yaml
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
+else
+	ifeq ($(TARGET), ocp)
+		cd config/default && $(KUSTOMIZE) edit add resource ../networking
+		cd config/networking && $(KUSTOMIZE) edit add resource route.yaml
+		cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+		$(KUSTOMIZE) build config/default | kubectl apply -f -
+		$(shell oc expose svc k4e-operator-controller-manager -n k4e-operator-system)
```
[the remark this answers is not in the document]
arielireni: removed

masayag (approved):

pkliczewski (approved):

config/manager/manager.yaml
```diff
@@ -66,26 +66,6 @@ spec:
...
       targetPort: 8043
-    - name: metrics
-      protocol: TCP
-      port: 8080
-      targetPort: 8080
   selector:
     control-plane: controller-manager
   type: ClusterIP
----
-kind: ServiceMonitor
```
jakub-dzon, line 77: Why is ServiceMonitor removed?

masayag: this resource doesn't include in k8s, at least not by default.
attempt to use it with existing code fails. therefore it would be better removing it and adding it to the docs and later we can think on adding a rule to deploy metrics.

jakub-dzon (approved):

eloycoto (approved):

PR #64, opened 2021-12-23
Data model for configuring on-device metrics retention

jakub-dzon:
This PR introduces per-device configuration options for metrics - data retention time and max size.


api/v1alpha1/edgedevice_types.go
```diff
@@ -36,6 +36,16 @@ type EdgeDeviceSpec struct {
...
 	Storage   *Storage                `json:"storage,omitempty"`
+	Metrics   *MetricsConfiguration   `json:"metrics,omitempty"`
+}
+
+type MetricsConfiguration struct {
+	Retention *Retention `json:"retention,omitempty"`
+}
+
+type Retention struct {
+	MaxMiB   int32 `json:"maxMiB"`
```
eloycoto, line 47: I would add here the description and the kubebuilder minimum if it's set.

eloycoto (approved):
LGTM, only a minor comment.

api/v1alpha1/edgedevice_types.go
jakub-dzon: Done.

masayag (approved):

PR #66, opened 2021-12-30
Add the namespace of the service account

masayag:

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ spec:
  ``` 
 -  Connecting Prometheus to our Custom Grafana
    -  Grant cluster-monitoring-view cluster role to the  grafana-serviceaccount service account, that was created alongside the Grafana instance\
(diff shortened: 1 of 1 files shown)


bardielle (approved):

bardielle:
https://docs.openshift.com/container-platform/3.7/admin_guide/manage_rbac.html#managing-role-bindings
`You can also manage cluster role bindings using the following operations. The -n flag is not used for these operations because cluster role bindings use non-namespaced resources.`

masayag:
> https://docs.openshift.com/container-platform/3.7/admin_guide/manage_rbac.html#managing-role-bindings `You can also manage cluster role bindings using the following operations. The -n flag is not used for these operations because cluster role bindings use non-namespaced resources.`

Without specifying the namespace there is a warning that the service account doesn't exist in the default namespace.

PR #70, opened 2022-01-10
Set limits and concurrency for k4e controller

masayag:
In order to prevent from having an impact on the cluster, setting limits
for the k4e controller.

In addition, to save memory, running the reconcilers with a higher
concurrency consume less memory than increasing number of replicas.

Memory and CPU request and limits need to be tuned based on expected
amount of devices.


main.go
```diff
@@ -115,6 +115,9 @@ var Config struct {
 
 	// Client Certificate expiration time
 	ClientCertExpirationTime uint `envconfig:"CLIENT_CERT_EXPIRATION_DAYS" default:"30"`
+
+	// MaxConcurrentReconciles is the maximum number of concurrent Reconciles which can be run
+	MaxConcurrentReconciles uint `envconfig:"MAX_CONCURRENT_RECONCILES" default:"3"`
```
gciavarrini, line 120: Question to learn: why is `MaxConcurrentReconciles` the same for edgedevice_controller, edgedeployment_controller and edgedevicelabels_controller?

jakub-dzon (approved):

gciavarrini (approved):

main.go
masayag: we can tune it as we go, if we find out a need for a different granularity for each.
the first idea was to process the same amount of requests as with 3 replicas.

pkliczewski: @masayag any reason?

pkliczewski (approved):

PR #71, opened 2022-01-12
Add int32 type for EdgeDeployment metrics port

masayag:

### DIFF
--- a/config/crd/bases/management.k4e.io_edgedeployments.yaml
+++ b/config/crd/bases/management.k4e.io_edgedeployments.yaml
@@ -125,6 +125,7 @@ spec:
                           type: string
                         port:
                           description: Port to use when retrieve the metrics
+                          format: int32
                           maximum: 65535
                           minimum: 0
                           type: integer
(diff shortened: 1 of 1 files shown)


config/crd/bases/management.k4e.io_edgedeployments.yaml
```diff
@@ -125,6 +125,7 @@ spec:
                           type: string
                         port:
                           description: Port to use when retrieve the metrics
+                          format: int32
```
gciavarrini, line 128: Question to learn: what were the problems due to the lack of the "format" field?

masayag: An annoying uncommitted change after each invocation of `make gen-manifests` :-) 
After each PR, we should expect the generate files not to be changed. The code itself already instructs the use of int32 (the pr was recently merged), but it misses this part.
About the functionality itself, the definition of the element in the CRD is:
```
                  port:
                    description: Port to use when retrieve the metrics
                    format: int32
                    maximum: 65535
                    minimum: 0
```
so the range would enforce the values. no harm should be caused without this PR.

gciavarrini (approved):

eloycoto (approved):

PR #72, opened 2022-01-12
Update metrics service

masayag:
- Move metrics port from http-server service to metrics-service.
- Restore metrics port.
- Add to operator's deployment pod's ports.
- Use port name instead of port number in service to map service port to pod's target port

### DIFF
--- a/config/rbac/auth_proxy_service.yaml
+++ b/config/rbac/auth_proxy_service.yaml
@@ -10,5 +10,9 @@ spec:
   - name: https
     port: 8443
     targetPort: https
+  - name: http
+    protocol: TCP
+    port: 8080
+    targetPort: metrics
   selector:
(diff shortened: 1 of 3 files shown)


config/rbac/auth_proxy_service.yaml
```diff
@@ -10,5 +10,9 @@ spec:
   - name: https
     port: 8443
     targetPort: https
+  - name: http
+    protocol: TCP
+    port: 8080
+    targetPort: 8080
```
jakub-dzon, line 16: WDYT about using port name reference instead of numeric value? If the pod port changes (yeah, unlikely), the service will not need to.

jakub-dzon (approved):

bardielle (approved):

PR #75, opened 2022-01-17
Added system metrics gathering interval

jakub-dzon:
This PR adds possibility to control device system metrics gathering interval.


pkliczewski (approved):

api/v1alpha1/edgedevice_types.go
```diff
@@ -40,7 +40,15 @@ type EdgeDeviceSpec struct {
...
-	Retention *Retention `json:"retention,omitempty"`
+	Retention     *Retention                  `json:"retention,omitempty"`
+	SystemMetrics *SystemMetricsConfiguration `json:"system,omitempty"`
+}
+
+type SystemMetricsConfiguration struct {
+	// Interval(in seconds) to scrape system metrics.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=60
+	Interval int32 `json:"interval,omitempty"`
```
gciavarrini, line 51: May I ask why `Interval` variables are defined as `int` in edgedeployment_type.go, `int64` in metrics.go, and `int32` here? IIUC, they all represents Interval in seconds.

jakub-dzon: Good point; I chose `int32` because I think it should be enough and not `int` to avoid conversion.

@eloycoto WDYT about that; are you ok with downsizing deployment one or do you think it should be `int64`? Gloria's right that should be the same type for edge device and edge deployment.

jakub-dzon: After all `int32` max value is a little over 68 years;) Maybe it's worth aligning to that type in #76?

gciavarrini: Also `WorkloadsMonitoringInterval` in device_configuration_message.go is `int64` (I know that's not related to this PR, just pointing out this :) )

eloycoto: All good, I'll change.

eloycoto (approved):

PR #78, opened 2022-01-17
EdgeDeployment pod spec validation

ydayagi:
Create validating admission webhook for EdgeDeployment
Based on
https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html

https://issues.redhat.com/browse/ECOPROJECT-217


jakub-dzon (commented):
Solid work!

main.go
```diff
@@ -219,6 +220,15 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "EdgeDeployment")
 		os.Exit(1)
 	}
+
+	// webhooks
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
```
gciavarrini, line 225: Shouldn't it be used in the same way as `OBC_AUTO_CREATE`?

gciavarrini:
WDYT, adding a subfolder `edgedeployment` in `v1alpha1`? Could it help to organize the code better?

README.md
```diff
@@ -5,6 +5,13 @@
 
 When the CRDs are present in the cluster, the operator can either be run outside or inside the cluster.
 
+## Prerequisites/Dependencies
+### cert-manager
+We use admission webhooks for managing the CRDs. Therefore, there's a need for TLS certificates and keys management.
```
pkliczewski, line 10: Would be good to add with which podman version we are aligned. For each new podman update we would need to make sure the validation is still correct.

api/v1alpha1/edgedeployment_webhook.go
```diff
@@ -0,0 +1,135 @@
...
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *EdgeDeployment) ValidateDelete() error {
+	edgedeploymentlog.Info("validate delete", "name", r.Name)
+
+	return nil
+}
+
+func (r *EdgeDeployment) validate() error {
+	var notValidPaths []string
```
eloycoto, line 77: I think that here instead of using an array of strings for errors, I would use go-multierror something that is already used in multiple places.

api/v1alpha1/edgedeployment_webhook_test.go
```diff
@@ -0,0 +1,174 @@
...
+import (
+	"github.com/jakub-dzon/k4e-operator/api/v1alpha1"
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+var _ = Describe("v1alpha1", func() {
```
eloycoto, line 12: I would change this name to edgeDeployment Webhook, so it's clear when something fails.

eloycoto (changes requested):
I think that we need to a way in the makefile to install cert-manager no?

README.md
ydayagi: we have not yet pinned a version. I think that podman is a key component so we should document it well somewhere else.

api/v1alpha1/edgedeployment_webhook.go
```diff
@@ -0,0 +1,135 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+//+kubebuilder:docs-gen:collapse=Go imports
+
+/*
+Next, we'll setup a logger for the webhooks.
+*/
+
+var edgedeploymentlog = logf.Log.WithName("edgedeployment-resource")
```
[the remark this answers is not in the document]
ydayagi: the controllers do the same. so i do not want t do something else. also.....there's no real need for this logger besides info messages. I'm not sure we should keep even these. what do u think?

api/v1alpha1/edgedeployment_webhook.go
ydayagi: I don't want 'validate' to determine how the error is produced. it only reports 'bad' paths. just like a parser (e.g. XML, JSON and so on). we may want to reuse the function.

ydayagi: the function returns an error. how the error is created is an internal matter. besides, GO gives me all the tools I need. and multierror is not appropriate. it is 'multi' and 'error'. here we have multi string.

api/v1alpha1/edgedeployment_webhook.go
```diff
@@ -0,0 +1,135 @@
...
+				} else if envVar.ValueFrom.ResourceFieldRef != nil {
+					notValidPaths = append(notValidPaths, envMsg(container, envVar, "resourceFieldRef"))
+				}
+			}
+		}
+	}
+
+	for _, volume := range podSpec.Volumes {
+		if volume.HostPath == nil {
+			notValidPaths = append(notValidPaths, "volumes can only contain hostPath volumes")
```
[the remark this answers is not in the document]
ydayagi: the message states: volume**s**, not volume, can only contain volume (single) of type hostPath.

api/v1alpha1/edgedeployment_webhook_test.go
ydayagi: r u refering to edgeDeployment variable?

api/v1alpha1/edgedeployment_webhook_test.go
```diff
@@ -0,0 +1,174 @@
...
+			// when
+			err := edgeDeployment.ValidateUpdate(nil)
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		table.DescribeTable("test all invalid fields", func(editEdgeDeployment func()) {
+			// given
+			editEdgeDeployment()
```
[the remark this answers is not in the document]
ydayagi: thank you

main.go
ydayagi: we don't want to put everything in config. OBC is indeed configuration. user may want to use his own S3 solution. However, webhook is not optional. it is only optional for development.

api/v1alpha1/edgedeployment_webhook.go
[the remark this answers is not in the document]
ydayagi: I will remove the message and add the path. I think it will be better cause it is consistent with the other validations and with the error message.

api/v1alpha1/edgedeployment_webhook.go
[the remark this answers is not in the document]
jakub-dzon: It was not intentional on our end - the controllers have been generated that way.
IMO logging in the webhook at the current state is not needed at all so I am all for removing it.

api/v1alpha1/edgedeployment_webhook_test.go
ydayagi: got it. changing

README.md
ydayagi: I added podman. please let me know if it's ok.

api/v1alpha1/edgedeployment_webhook.go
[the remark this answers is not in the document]
ydayagi: ok. will do. thank you all.

main.go
gciavarrini: So are you trying to prevent disabling webhook in prod environment in this way?

ydayagi:
> I think that we need to a way in the makefile to install cert-manager no?

it's like OCS. it's ok tolet user do it. I documented as prerequisite in readme.
we do not have a specific version we require, and more importantly, it is a process that takes time and may fail. so if we make it part of deploy, then deploy will fail. I tried :-)

main.go
ydayagi: yes

eloycoto:
>it's like OCS. it's ok tolet user do it. I documented as prerequisite in readme.
we do not have a specific version we require, and more importantly, it is a process that takes time and may fail. so if we make it part of deploy, then deploy will fail. I tried :-)


But at some point, we need to have all this install process in the makefile, scripts or whatever for the CI, it's something that we're not testing right now, and we need.

api/v1alpha1/edgedeployment_webhook.go
eloycoto: +1

ydayagi:
> > it's like OCS. it's ok tolet user do it. I documented as prerequisite in readme.
> > we do not have a specific version we require, and more importantly, it is a process that takes time and may fail. so if we make it part of deploy, then deploy will fail. I tried :-)
> 
> But at some point, we need to have all this install process in the makefile, scripts or whatever for the CI, it's something that we're not testing right now, and we need.

could be. only time will tell. but at the moment our dependencies are installed manually. Let's keep in mind that we're talking about a single YAML to apply. So it's not like that a make target helps you. For CI/e2e we can add it as a step to the cluster setup. Not sure that we'll want to use it for CI, though.

eloycoto:
>could be. only time will tell. but at the moment our dependencies are installed manually. Let's keep in mind that we're talking about a single YAML to apply. So it's not like that a make target helps you. For CI/e2e we can add it as a step to the cluster setup. Not sure that we'll want to use it for CI, though.

@machacekondra any feedback here?

api/v1alpha1/edgedeployment_webhook.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	if len(notValidPaths) != 0 {
+		return errors.New("the following paths in podSpec are not valid: " +
```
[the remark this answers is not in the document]
ydayagi: @jakub-dzon please review my fixes

api/v1alpha1/edgedeployment_webhook.go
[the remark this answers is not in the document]
ydayagi: @gciavarrini  please review my fixes

README.md
ydayagi: @pkliczewski please review my fixes

README.md
```diff
@@ -5,6 +5,17 @@
 
 When the CRDs are present in the cluster, the operator can either be run outside or inside the cluster.
 
+## Prerequisites/Dependencies
+### podman
+This project relies strongly on Podman. It is used in the edge devices and also in the operator.
```
masayag, line 10: Could you clarify what's the meaning of "podman...also in the operator" ?

config/webhook/kustomizeconfig.yaml
```diff
@@ -0,0 +1,25 @@
+# the following config is for teaching kustomize where to look at when substituting vars.
+# It requires kustomize v2.1.0 or newer to work properly.
+nameReference:
+- kind: Service
+  version: v1
+  fieldSpecs:
+  - kind: MutatingWebhookConfiguration
```
masayag, line 7: Can we omit the MutatingWebhookConfiguration? we don't handle it.

config/webhook/kustomizeconfig.yaml
```diff
@@ -0,0 +1,25 @@
...
+    group: admissionregistration.k8s.io
+    path: webhooks/clientConfig/service/name
+  - kind: ValidatingWebhookConfiguration
+    group: admissionregistration.k8s.io
+    path: webhooks/clientConfig/service/name
+
+namespace:
+- kind: MutatingWebhookConfiguration
```
masayag, line 15: same here

README.md
ydayagi: you and I know we do not use podman directly in the operator. at least not at the moment (see ). But operator does rely on poman version for validation. That' what I meant. If you have a different suggestion I'll be happy to use it. I was just trying to make sure users/developers have all the info.

config/webhook/kustomizeconfig.yaml
ydayagi: Correct. We can. I was waiting for your (all reviewers) feedback. Will remove.

config/webhook/kustomizeconfig.yaml
ydayagi: sure

jakub-dzon (approved):

pkliczewski (approved):
Looks good to me let's make sure others are happy as wll.

masayag (approved):

PR #79, opened 2022-01-19
Enable monitoring for OCP

masayag:
The patch adds the missing resources to enable monitoring for k4e.
In addition, in order to avoid from collision due to services with the
same labels/selector and the same port name (https), the http server
ports were renamed to yggd and yddgs.
Alternately, the https port name of the metrics can be renamed.
Before this PR, the endpoints of two services was the same, while there
was no representation for the :8043 port.

```
→ oc get ep -n k4e-operator-system
NAME                                              ENDPOINTS                             AGE
k4e-operator-controller-manager                   172.30.0.151:8888,172.30.0.151:8443   2m10s
k4e-operator-controller-manager-metrics-service   172.30.0.151:8443,172.30.0.151:8080   2m10s
```

With this PR, all 4 ports are presented, according to the service:
```
→ oc get ep -n k4e-operator-system -o wide
NAME                                              ENDPOINTS                             AGE
k4e-operator-controller-manager                   172.30.0.158:8043,172.30.0.158:8888   50s
k4e-operator-controller-manager-metrics-service   172.30.0.158:8443,172.30.0.158:8080   50s

→ oc get svc -n k4e-operator-system -o wide
NAME                                              TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)             AGE   SELECTOR
k4e-operator-controller-manager                   ClusterIP   10.128.171.232   <none>        8888/TCP,8043/TCP   61s   control-plane=controller-manager
k4e-operator-controller-manager-metrics-service   ClusterIP   10.128.104.37    <none>        8443/TCP,8080/TCP   61s   control-plane=controller-manager
```

This PR doesn't include the update required for enabling monitoring in OCP which can be done by running:
```
kubectl apply -f - <<EOF
apiVersion: v1
data:
  config.yaml: |
    enableUserWorkload: true
kind: ConfigMap
metadata:
  name: cluster-monitoring-config
  namespace: openshift-monitoring
EOF
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
+    openshift.io/cluster-monitoring: "true"
   name: system
 ---
 apiVersion: apps/v1
@@ -34,9 +35,9 @@ spec:
         name: manager
         ports:
           - containerPort: 8888
-            name: http
+            name: yggd
           - containerPort: 8043
-            name: https
+            name: yggds
           - containerPort: 8080
             name: metrics
         securityContext:
@@ -72,14 +73,14 @@ metadata:
     control-plane: controller-manager
 spec:
   ports:
-    - name: http
+    - name: yggd
       protocol: TCP
       port: 8888
-      targetPort: http
-    - name: https
+      targetPort: yggd
+    - name: yggds
       protocol: TCP
       port: 8043
-      targetPort: https
+      targetPort: yggds
   selector:
     control-plane: controller-manager
   type: ClusterIP
--- a/config/prometheus/prometheus_role_binding.yaml
+++ b/config/prometheus/prometheus_role_binding.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s
+  namespace: system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: openshift-monitoring
--- a/docs/metrics/metrics.md
+++ b/docs/metrics/metrics.md
@@ -1,39 +1,16 @@
-# Publication of the metrics
+# Publication of the metrics in Prometheus on OCP
 
 In order to publish the metrics several steps need to be done:
-- Create a service monitor
-```yaml
-apiVersion: monitoring.coreos.com/v1
-kind: ServiceMonitor
-metadata:
-  name: k4e-operator-servicemonitor
-  namespace: k4e-operator-system
-  labels:
-    control-plane: controller-manager
-spec:
-  endpoints:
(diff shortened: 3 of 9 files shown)


bardielle (approved):

config/manager/manager.yaml
```diff
@@ -34,9 +35,9 @@ spec:
         name: manager
         ports:
           - containerPort: 8888
-            name: http
+            name: yggd
```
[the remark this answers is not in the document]
jakub-dzon: Good idea! I like it.

docs/metrics/metrics.md
```diff
@@ -1,39 +1,16 @@
...
-data:
-  config.yaml: |
-    enableUserWorkload: true
 kind: ConfigMap
 metadata:
   name: cluster-monitoring-config
   namespace: openshift-monitoring
+data:
+  config.yaml: |
+  EOF
```
jakub-dzon, line 15: Isn't `enableUserWorkload: true` needed there?

config/prometheus/prometheus_role_binding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: prometheus-k8s
+  namespace: system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: openshift-monitoring
```
machacekondra, line 13: Just wonder whether we care about monitoring on k8s? in that case this should be configurable.

docs/metrics/metrics.md
```diff
@@ -1,39 +1,16 @@
...
-- Enable user-defined for user projects
-```yaml
+
+- Enable monitoring on the cluster
+```bash
+kubectl apply -f - <<EOF
 apiVersion: v1
-data:
-  config.yaml: |
-    enableUserWorkload: true
```
machacekondra, line 34: I think long-term we should think about producing the metrics to default prometheus, instead of prometheus for userworkload.

machacekondra (approved):

docs/metrics/metrics.md
masayag: It isn't needed based on the admin guide and as verified by testing it.

https://docs.openshift.com/container-platform/4.8/monitoring/configuring-the-monitoring-stack.html#creating-cluster-monitoring-configmap_configuring-the-monitoring-stack

config/prometheus/prometheus_role_binding.yaml
masayag: we'll have to address k8s in a different task.
for that purpose we'd probably like to change the way we're using kustomize.
also, for k8s we should ease the process of installing prometheus.
but that is out of the boundaries of this PR/task.

docs/metrics/metrics.md
jakub-dzon: Thanks! Interesting... I was following https://docs.openshift.com/container-platform/4.8/monitoring/enabling-monitoring-for-user-defined-projects.html but it seems to be inaccurate.

jakub-dzon (approved):

PR #81, opened 2022-01-20
Allowing user to disable system metrics collection

jakub-dzon:
This PR allows user to disable system metrics collection from a device.

### DIFF
--- a/api/v1alpha1/edgedevice_types.go
+++ b/api/v1alpha1/edgedevice_types.go
@@ -53,6 +53,9 @@ type SystemMetricsConfiguration struct {
 	// AllowList defines name and namespace of a ConfigMap containing
 	// list of system metrics that should be scraped
 	AllowList *ObjectRef `json:"allowList,omitempty"`
+
+	// Disabled when set to true instructs the device to turn off system metrics collection
+	Disabled bool `json:"disabled,omitempty"`
 }
 
 type Retention struct {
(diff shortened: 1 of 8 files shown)


ydayagi:
I wonder, should we turn off node_exporter? For the hardware collection, we were concerned about the collection of redundant/unused information. We worried that it takes some more CPU. Is it because node_exporter is a separate process that we do not worry?

jakub-dzon:
Yes! I plan to turn off node_exporter service if disabled=true.

masayag (approved):

api/v1alpha1/edgedevice_types.go
```diff
@@ -53,6 +53,9 @@ type SystemMetricsConfiguration struct {
 	// AllowList defines name and namespace of a ConfigMap containing
 	// list of system metrics that should be scraped
 	AllowList *ObjectRef `json:"allowList,omitempty"`
+
+	// Disabled when set to true instructs the device to turn off system metrics collection
+	Disabled bool `json:"disabled,omitempty"`
```
machacekondra, line 58: Just out of curiosity why using `disable` instead of `enable`? I think mostly there are `enable` options instead of `disable`.

machacekondra (approved):

api/v1alpha1/edgedevice_types.go
jakub-dzon: Metrics are enabled by default and `enabled` handling would be more cumbersome than `disabled`.

pkliczewski (approved):

PR #82, opened 2022-01-20
Custom labels for hardware

pkliczewski:
We would like to use labels with hardware information to select devices where to run workload.

Here is how the labels look like after this change:

```yaml
    labels:
      device.cpu-architecture: x86_64
      device.cpu-model: intelrcoretmi7-8665ucpu1.90ghz
      device.hostname: localhost.localdomain
      device.system-manufacturer: lenovo
      device.system-product: 20n5s2nc1v
      device.system-serial: PF20YKWG
```

Signed-off-by: Piotr Kliczewski <piotr.kliczewski@gmail.com>

### DIFF
--- a/internal/hardware/hardware.go
+++ b/internal/hardware/hardware.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/jakub-dzon/k4e-operator/api/v1alpha1"
+	"github.com/jakub-dzon/k4e-operator/internal/utils"
 	"github.com/jakub-dzon/k4e-operator/models"
 )
 
@@ -88,4 +89,46 @@ func MapHardware(hardware *models.HardwareInfo) *v1alpha1.Hardware {
 		}
 	}
 	return &hw
+}
+
+func MapLabels(hardware *models.HardwareInfo) map[string]string {
+	if hardware == nil {
+		return nil
+	}
+	var labels = make(map[string]string)
+
+	hostname, err := utils.NormalizeLabel(hardware.Hostname)
+	if err == nil {
+		labels["device.hostname"] = hostname
+	}
+
+	cpu := hardware.CPU
+	if cpu != nil {
+		arch, err := utils.NormalizeLabel(cpu.Architecture)
+		if err == nil {
+			labels["device.cpu-architecture"] = arch
+		}
+		model, err := utils.NormalizeLabel(cpu.ModelName)
+		if err == nil {
+			labels["device.cpu-model"] = model
+		}
+	}
+
+	systemVendor := hardware.SystemVendor
+	if systemVendor != nil {
+		manufacturer, err := utils.NormalizeLabel(systemVendor.Manufacturer)
+		if err == nil {
+			labels["device.system-manufacturer"] = manufacturer
+		}
+		productName, err := utils.NormalizeLabel(systemVendor.ProductName)
+		if err == nil {
+			labels["device.system-product"] = productName
+		}
+		serialNumber, err := utils.NormalizeLabel(systemVendor.SerialNumber)
+		if err == nil {
+			labels["device.system-serial"] = serialNumber
+		}
+	}
+
+	return labels
 }
--- a/internal/repository/edgedevice/edgedevice.go
+++ b/internal/repository/edgedevice/edgedevice.go
@@ -2,6 +2,8 @@ package edgedevice
 
 import (
 	"context"
+	"reflect"
+	"time"
 
 	"github.com/jakub-dzon/k4e-operator/api/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -16,6 +18,7 @@ type Repository interface {
 	Patch(ctx context.Context, old, new *v1alpha1.EdgeDevice) error
 	ListForSelector(ctx context.Context, selector *metav1.LabelSelector, namespace string) ([]v1alpha1.EdgeDevice, error)
 	RemoveFinalizer(ctx context.Context, edgeDevice *v1alpha1.EdgeDevice, finalizer string) error
+	UpdateLabels(ctx context.Context, device *v1alpha1.EdgeDevice, labels map[string]string) error
 }
 
 type CRRepository struct {
@@ -80,4 +83,45 @@ func (r *CRRepository) RemoveFinalizer(ctx context.Context, edgeDevice *v1alpha1
 	}
 
 	return nil
+}
+
+func (r *CRRepository) UpdateLabels(ctx context.Context, device *v1alpha1.EdgeDevice, labels map[string]string) error {
+	err := r.updateLabels(ctx, device, labels)
+	if err == nil {
+		return nil
+	}
+
+	// retry patching the edge device labels because the device can be update concurrently
+	for i := 1; i < 4; i++ {
+		time.Sleep(time.Duration(i*50) * time.Millisecond)
+		device2, err := r.Read(ctx, device.Name, device.Namespace)
+		if err != nil {
+			continue
+		}
+		err = r.updateLabels(ctx, device2, labels)
+		if err == nil {
+			return nil
+		}
+	}
+	return err
+}
+
+func (r *CRRepository) updateLabels(ctx context.Context, device *v1alpha1.EdgeDevice, labels map[string]string) error {
+	deviceCopy := device.DeepCopy()
+	deviceLabels := deviceCopy.Labels
+	if deviceLabels == nil {
+		deviceLabels = make(map[string]string)
+	}
+	for key, value := range labels {
+		deviceLabels[key] = value
+	}
+	if reflect.DeepEqual(deviceLabels, device.Labels) {
+		return nil
+	}
+	deviceCopy.Labels = deviceLabels
+	err := r.Patch(ctx, device, deviceCopy)
+	if err == nil {
+		device.Labels = deviceCopy.Labels
+	}
+	return err
 }
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -1,6 +1,11 @@
 package utils
 
 import (
+	"fmt"
+	"regexp"
+	"strings"
+	"unicode"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -12,4 +17,24 @@ func HasFinalizer(cr *metav1.ObjectMeta, name string) bool {
 		}
 	}
 	return false
+}
+
+// NormalizeLabel normalizes label string according to k8s format
+func NormalizeLabel(name string) (string, error) {
+	// convert name to lowercase
+	name = strings.ToLower(name)
+
+	// slice string based on first and last alphanumeric character
+	firstLegal := strings.IndexFunc(name, func(c rune) bool { return unicode.IsLower(c) || unicode.IsDigit(c) })
+	lastLegal := strings.LastIndexFunc(name, func(c rune) bool { return unicode.IsLower(c) || unicode.IsDigit(c) })
+
+	if firstLegal < 0 {
+		return "", fmt.Errorf("The name doesn't contain a legal alphanumeric character")
+	}
+
+	name = name[firstLegal : lastLegal+1]
+	reg := regexp.MustCompile("[^a-z0-9-_.]+")
+	name = reg.ReplaceAllString(name, "")
+
+	return name, nil
 }
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
 	"github.com/jakub-dzon/k4e-operator/internal/devicemetrics"
 	"github.com/jakub-dzon/k4e-operator/internal/heartbeat"
 
@@ -335,6 +336,12 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 
 		if err != nil {
 			logger.Error(err, "cannot update EdgeDevice status")
+			h.metrics.IncEdgeDeviceFailedRegistration()
+			return operations.NewPostDataMessageForDeviceInternalServerError()
+		}
+		err = h.deviceRepository.UpdateLabels(ctx, &device, hardware.MapLabels(registrationInfo.Hardware))
+		if err != nil {
+			logger.Error(err, "cannot update EdgeDevice labels")
 			h.metrics.IncEdgeDeviceFailedRegistration()
 			return operations.NewPostDataMessageForDeviceInternalServerError()
 		}
--- a/internal/hardware/hardware_test.go
+++ b/internal/hardware/hardware_test.go
@@ -1,6 +1,8 @@
 package hardware_test
 
 import (
+	"strings"
+
 	"github.com/jakub-dzon/k4e-operator/api/v1alpha1"
 	"github.com/jakub-dzon/k4e-operator/internal/hardware"
 	"github.com/jakub-dzon/k4e-operator/models"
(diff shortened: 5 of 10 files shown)


internal/hardware/hardware.go
```diff
@@ -89,3 +89,32 @@ func MapHardware(hardware *models.HardwareInfo) *v1alpha1.Hardware {
...
 	return &hw
 }
+
+func MapLabels(hardware *models.HardwareInfo) map[string]string {
+	if hardware == nil {
+		return nil
+	}
+	var labels = make(map[string]string)
+
+	for _, g := range hardware.Gpus {
```
jakub-dzon, line 99: GPU information is no longer sent after https://github.com/jakub-dzon/k4e-device-worker/pull/73.

internal/yggdrasil/yggdrasil.go
```diff
@@ -392,6 +404,39 @@ func (h *Handler) updateDeviceStatus(ctx context.Context, device *v1alpha1.EdgeD
 	return err
 }
 
+func (h *Handler) updateDeviceLabels(ctx context.Context, device *v1alpha1.EdgeDevice, labels map[string]string) error {
```
jakub-dzon, line 407: WDYT about moving this code to the EdgeDevice repository?

jakub-dzon (changes requested):

internal/hardware/hardware.go
pkliczewski: OK, will remove

internal/yggdrasil/yggdrasil.go
pkliczewski: sure, will move

gciavarrini (approved):
lgtm

internal/repository/edgedevice/edgedevice.go
```diff
@@ -81,3 +83,36 @@ func (r *CRRepository) RemoveFinalizer(ctx context.Context, edgeDevice *v1alpha1
...
+func (r *CRRepository) updateLabels(ctx context.Context, device *v1alpha1.EdgeDevice, labels map[string]string) error {
+	deviceCopy := device.DeepCopy()
+	deviceLabels := deviceCopy.ObjectMeta.GetLabels()
+	if deviceLabels == nil {
+		deviceLabels = make(map[string]string)
+	}
+	for key, value := range labels {
+		deviceLabels[key] = value
+	}
+	return r.Patch(ctx, device, deviceCopy)
```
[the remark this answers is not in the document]
pkliczewski: My assumption was that Patch method is checking it. I will double check this and fix if needed.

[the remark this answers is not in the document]
pkliczewski: I checked and was not able to find it. Will push update. Thank you for good catch!

jakub-dzon:
LGTM. 
Why is this PR WIP? What is missing?

pkliczewski:
> LGTM. Why is this PR WIP? What is missing?

I need to run it to confirm it works. I need to setup the env for this.

internal/yggdrasil/yggdrasil.go
```diff
@@ -313,6 +314,12 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 		if err != nil {
 			return operations.NewPostDataMessageForDeviceInternalServerError()
 		}
+		if heartbeat.Hardware != nil {
+			err = h.deviceRepository.UpdateLabels(ctx, edgeDevice, hardware.MapLabels(heartbeat.Hardware))
```
masayag, line 318: will this cause labels changes performed by the admin be overridden by next heartbeat?

internal/yggdrasil/yggdrasil.go
```diff
@@ -357,6 +364,12 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 			h.metrics.IncEdgeDeviceFailedRegistration()
 			return operations.NewPostDataMessageForDeviceInternalServerError()
 		}
+		err = h.deviceRepository.UpdateLabels(ctx, &device, hardware.MapLabels(registrationInfo.Hardware))
```
masayag, line 367: this shouldn't produce too much load (expected to be called once) and yet, could update the labels as part of the device creation or is the order(create device --> update status --> update labels) important here?

internal/hardware/hardware.go
```diff
@@ -89,3 +89,27 @@ func MapHardware(hardware *models.HardwareInfo) *v1alpha1.Hardware {
...
+	if hardware == nil {
+		return nil
+	}
+	var labels = make(map[string]string)
+
+	labels["feature.device/hostname"] = hardware.Hostname
+
+	cpu := hardware.CPU
+	if cpu != nil {
+		labels[fmt.Sprintf("feature.device/cpu.%s", cpu.Architecture)] = "true"
```
masayag, line 103: can we assure the value of `cpu.Architecture` is in lowercase? (labels should be in lower-case)

internal/hardware/hardware.go
```diff
@@ -89,3 +89,27 @@ func MapHardware(hardware *models.HardwareInfo) *v1alpha1.Hardware {
...
+		labels[fmt.Sprintf("feature.device/cpu.%s", cpu.ModelName)] = "true"
```
masayag, line 104: can we assure the value of `cpu.ModelName` is in lowercase?

internal/hardware/hardware.go
```diff
@@ -89,3 +89,27 @@ func MapHardware(hardware *models.HardwareInfo) *v1alpha1.Hardware {
...
+	}
+
+	systemVendor := hardware.SystemVendor
+	if systemVendor != nil {
+		labels[fmt.Sprintf("feature.device/system.manufacturer.%s", systemVendor.Manufacturer)] = "true"
```
masayag, line 109: same here

internal/hardware/hardware.go
masayag, line 103: Was there a reason for selecting a label name to start with `feature.device/`?
If there is no strong reason for it, perhaps we should have a common prefix for all of operator's labels (and annotation)

pkliczewski: I tried to follow convention used [here](https://github.com/kubernetes-sigs/node-feature-discovery). We can change it. It is up to us to define common prefix.

internal/yggdrasil/yggdrasil.go
pkliczewski: From my point of view the order is not important. Do you think we should change it?

internal/hardware/hardware.go
pkliczewski: Agreed, we should not assume lowercase and do it. Thank you for pointing this!

internal/yggdrasil/yggdrasil.go
pkliczewski: If there is the same key used it will override the value but other labels will be untouched.

internal/repository/edgedevice/edgedevice.go
```diff
@@ -81,3 +84,39 @@ func (r *CRRepository) RemoveFinalizer(ctx context.Context, edgeDevice *v1alpha1
...
 	return nil
 }
+
+func (r *CRRepository) UpdateLabels(ctx context.Context, device *v1alpha1.EdgeDevice, labels map[string]string) error {
+	err := r.updateLabels(ctx, device, labels)
+	if err == nil {
+		return nil
+	}
+
+	// retry patching the edge device labels
```
machacekondra, line 94: Could you please add a comment why we need to retry here?

machacekondra (approved):

internal/repository/edgedevice/edgedevice.go
pkliczewski: Sure, will add

jakub-dzon (approved):

internal/hardware/hardware.go
pkliczewski: For some reason k8s complains about `/`. I updated the prefix. Please let me know whether it is something we would like to use.

pkliczewski:
I am not happy how the labels looks like after normalization:

>      feature.device-cpu.intelrcoretmi7-8665ucpu1-90ghz: "true"
>      feature.device-cpu.x8664: "true"

Any ideas how to improve it?

internal/utils/utils.go
```diff
@@ -13,3 +18,24 @@ func HasFinalizer(cr *metav1.ObjectMeta, name string) bool {
 	}
 	return false
 }
+
+// NormalizeLabel normalizes label string according to k8s format
+func NormalizeLabel(name string) (string, error) {
+	// convert name to lowercase and replace '.' with '-'
+	name = strings.ToLower(name)
+	name = strings.Replace(name, ".", "-", -1)
```
ydayagi, line 26: isn't '.' is valid in label name?

internal/utils/utils.go
```diff
@@ -13,3 +18,24 @@ func HasFinalizer(cr *metav1.ObjectMeta, name string) bool {
 	}
 	return false
 }
+
+// NormalizeLabel normalizes label string according to k8s format
+func NormalizeLabel(name string) (string, error) {
+	// convert name to lowercase and replace '.' with '-'
+	name = strings.ToLower(name)
```
ydayagi, line 25: I think that upper case is allowed in label names

internal/yggdrasil/yggdrasil.go
ydayagi: I think it should be in reconcile. Besides status (that has to be here, sort of) we do all such operations in reconcile. 
A client sends a request and should get a response asap. Less load on HTTP server. Think of many heartbeats.
Another motivation is that every step s another possible failure for the registration/heartbeat. I don't think it should fail it.

internal/hardware/hardware.go
```diff
@@ -89,3 +90,39 @@ func MapHardware(hardware *models.HardwareInfo) *v1alpha1.Hardware {
...
+	}
+	var labels = make(map[string]string)
+
+	labels["feature.device-hostname"] = hardware.Hostname
+
+	cpu := hardware.CPU
+	if cpu != nil {
+		arch, err := utils.NormalizeLabel(cpu.Architecture)
+		if err == nil {
+			labels[fmt.Sprintf("feature.device-cpu.%s", arch)] = "true"
```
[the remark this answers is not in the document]
pkliczewski: We were suggested to use [node-feature-discovery](https://github.com/kubernetes-sigs/node-feature-discovery) which uses this pattern. I think we need to reconsider and use it as values.

internal/hardware/hardware.go
```diff
@@ -89,3 +90,39 @@ func MapHardware(hardware *models.HardwareInfo) *v1alpha1.Hardware {
...
 	return &hw
 }
+
+func MapLabels(hardware *models.HardwareInfo) map[string]string {
+	if hardware == nil {
+		return nil
+	}
+	var labels = make(map[string]string)
+
+	labels["feature.device-hostname"] = hardware.Hostname
```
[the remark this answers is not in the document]
pkliczewski: Same. The pattern is taken from `node-feature-discovery`

internal/utils/utils.go
```diff
@@ -13,3 +18,24 @@ func HasFinalizer(cr *metav1.ObjectMeta, name string) bool {
 	}
 	return false
 }
+
+// NormalizeLabel normalizes label string according to k8s format
```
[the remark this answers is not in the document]
pkliczewski: It is existing func. We used it in the past but unfortunately project was archived so I haven't vendor it.

internal/utils/utils.go
pkliczewski: You are correct. It is allowed `([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')` but it seems that all the projects that I am aware are lowercasing the keys.

internal/utils/utils.go
pkliczewski: There is nothing in the docs about `:` being not allowed.

pkliczewski:
@ydayagi I am not sure what mean in this [comment](https://github.com/jakub-dzon/k4e-operator/pull/82#discussion_r790236893). For some reason I am not able to reply directly.

The code for processing hardware info is processed based on our rest api calls (registration and heartbeat). Please suggest how would change it.

internal/yggdrasil/yggdrasil.go
ydayagi: @pkliczewski see my comment to this thread

internal/hardware/hardware.go
ydayagi: according to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ it can be upper case. If use the value and not the name then it can be anything and we can use the original string

ydayagi: @pkliczewski

internal/hardware/hardware.go
ydayagi: @pkliczewski
according to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ it can be upper case. If use the value and not the name then it can be anything and we can use the original string

internal/hardware/hardware.go
ydayagi: @pkliczewski I do not understand what is the reason to follow what others are doing. We know what to do and what makes sense. I don't see any reasoning for this pattern.

internal/yggdrasil/yggdrasil.go
ydayagi: when heartbeat patches the status 'EdgeDeviceReconciler.Reconcile' is called. you can verify there that the hardware labels match the hardware status. For registration, it is not patch. it is create. still, it gets to Reconcile.

internal/hardware/hardware.go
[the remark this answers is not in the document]
ydayagi: @pkliczewski I do not know what others are doing. I do know what k8s gives me (key-value pairs called labels, label queries). I don't see any other way to go. If someone is doing it differently then...OMG

internal/hardware/hardware.go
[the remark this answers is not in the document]
ydayagi: well....their name is 'node-**feature**-discovery'. We do not have 'feature' in our project.

internal/utils/utils.go
ydayagi: I meant that dot '.' is valid. As well as all the other characters we are trying to replace.

internal/utils/utils.go
ydayagi: So?

internal/utils/utils.go
pkliczewski: ack, fixed

internal/hardware/hardware.go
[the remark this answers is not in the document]
pkliczewski: updated

internal/hardware/hardware.go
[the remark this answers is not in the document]
pkliczewski: fixed the label name pattern

pkliczewski:
@ydayagi please take a look

internal/utils/utils.go
```diff
@@ -13,3 +18,23 @@ func HasFinalizer(cr *metav1.ObjectMeta, name string) bool {
...
+func NormalizeLabel(name string) (string, error) {
```
jakub-dzon, line 23: Please add tests.

internal/yggdrasil/yggdrasil.go
```diff
@@ -314,6 +315,12 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 		if err != nil {
 			return operations.NewPostDataMessageForDeviceInternalServerError()
 		}
+		if heartbeat.Hardware != nil {
+			err = h.deviceRepository.UpdateLabels(ctx, edgeDevice, hardware.MapLabels(heartbeat.Hardware))
```
jakub-dzon, line 319: Please make sure you've got the most recent changes and move that call to https://github.com/jakub-dzon/k4e-operator/blob/main/internal/heartbeat/synchronous-handler.go

jakub-dzon (changes requested):

jakub-dzon (approved):

PR #84, opened 2022-01-25
Add configmap support

machacekondra:
This PR add support to manage configmaps.

We analaze the edgedeployments specification for the notion of the configmaps and we fetch those configmaps from k8s API.
Then we copy the configmap, remove unnecessary metadata and pass it to to k4-device-worker.


internal/utils/utils.go
```diff
@@ -1,9 +1,13 @@
 package utils
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type KeyMapType = map[string]interface{}
```
jakub-dzon, line 8: WDYT about `Set` ?

internal/utils/utils.go
```diff
@@ -13,3 +17,80 @@ func HasFinalizer(cr *metav1.ObjectMeta, name string) bool {
 	}
 	return false
 }
+
+func ExtractInfoFromEnvFrom(envs []corev1.EnvFromSource, themap MapType, ref func(interface{}) (bool, *bool, string)) {
```
jakub-dzon, line 21: WDYT about aliasing `func(interface{}) (bool, *bool, string)` so it was clear what it is supposed to do?

internal/utils/utils.go
```diff
@@ -13,3 +17,80 @@ func HasFinalizer(cr *metav1.ObjectMeta, name string) bool {
...
+			if optional {
+				themap[name] = nil
+			} else {
+				themap[name] = KeyMapType{}
+			}
+		}
+	}
+}
+
+func ExtractInfoFromEnv(env []corev1.EnvVar, configmapMap MapType, ref func(env corev1.EnvVar) (bool, *bool, string, string)) {
```
jakub-dzon, line 69: Alias for `func(env corev1.EnvVar) (bool, *bool, string, string)` ?

internal/utils/utils.go
```diff
@@ -13,3 +17,80 @@ func HasFinalizer(cr *metav1.ObjectMeta, name string) bool {
...
+			if optional {
+				themap[name] = nil
+			} else {
+				themap[name] = KeyMapType{}
+			}
+		}
+	}
+}
+
+func ExtractInfoFromVolume(volumes []corev1.Volume, themap MapType, ref func(interface{}) (bool, *bool, string)) {
```
jakub-dzon, line 45: Alias for `func(interface{}) (bool, *bool, string)`

internal/utils/utils.go
jakub-dzon, line 21: Please add tests for the new `ExtractInfoFrom*` methods.

internal/configmaps/configmaps.go
```diff
@@ -0,0 +1,109 @@
...
+
+type configMap struct {
+	client k8sclient.K8sClient
+}
+
+func NewConfigMap(client k8sclient.K8sClient) ConfigMap {
+	return &configMap{client: client}
+}
+
+func (cm *configMap) Fetch(ctx context.Context, deployments []v1alpha1.EdgeDeployment, device *v1alpha1.EdgeDevice) (models.ConfigmapList, error) {
```
jakub-dzon, line 31: Please add tests.

jakub-dzon (changes requested):

internal/utils/utils.go
machacekondra: What exactly do you mean here?

jakub-dzon: I mean `StringSet` would be more descriptive name of the data structure than `KeyMapType`

internal/configmaps/configmaps.go
```diff
@@ -0,0 +1,109 @@
...
+				return false, nil, ""
+			})
+		}
+	}
+
+	// read configmaps and add to configmaps list
+	for name, keys := range cmMap {
+		configmapObj, err := cm.readAndValidateConfigMap(ctx, name, namespace, keys)
+		if err != nil {
+			return nil, err
```
eloycoto, line 57: should we add some context to this err message no? Like name/namespace.

internal/configmaps/configmaps.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		if configmapObj == nil {
+			continue
+		}
+		configmapObj.ObjectMeta = metav1.ObjectMeta{Name: name}
```
eloycoto, line 62: the namespace should be added, no? I mean, you return the list, there is no way to get the same name?

internal/configmaps/configmaps_test.go
```diff
@@ -0,0 +1,205 @@
...
+								LocalObjectReference: v1.LocalObjectReference{
+									Name: "mycm1",
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+		deployments := []v1alpha1.EdgeDeployment{{
```
eloycoto, line 180: I would create a new helper function getDeplyment, and podData as argument.

eloycoto (approved):

internal/configmaps/configmaps.go
machacekondra: We don't need the namespace in the podman.

pkliczewski: in this case how do we handle name collisions? Once approval process will be implemented we will support many namespaces.

internal/configmaps/configmaps.go
```diff
@@ -0,0 +1,107 @@
...
+			return nil, err
+		}
+		list = append(list, string(obj))
+	}
+
+	return list, nil
+}
+
+func (cm *configMap) readAndValidateConfigMap(ctx context.Context, configmapName, configmapNamespace string, configmapKeys utils.StringSet) (*corev1.ConfigMap, error) {
+	optional := configmapKeys == nil
```
pkliczewski, line 72: We use this variable once. Why not to use this check in if (line 76)?

internal/configmaps/configmaps.go
machacekondra: The only way we could solve it is to change the name with the prefix with namespace.
So the ConfigMap in podman would be for example:
```yaml
kind: ConfigMap
metadata:
  name: configmapnamespace_configmapname
```

But we would need to in that case modify the podspec definition of the pod on device side, because the `configMapRef` would not be the same as k8s spec. In kubernetes world user can use only configmaps from same namespace as pod is running. That's why there is no specification of configMapRef::namespace. So maybe the concept of using only configmaps which are in same namespace as device is meaningfull here.

eloycoto (approved):

internal/configmaps/configmaps.go
jakub-dzon: AFAIR in case of Pods ConfigMap has to be in the same namespace as the Pod; I don't think we need to diverge from that and that implies that collisions would be impossible.

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
+				if env.ValueFrom.ConfigMapKeyRef != nil {
+					return true, env.ValueFrom.ConfigMapKeyRef.Optional, env.ValueFrom.ConfigMapKeyRef.Name, env.ValueFrom.ConfigMapKeyRef.Key
+				}
+				return false, nil, "", ""
+			})
+
+			// when
+			_, ok := maptypes["cm1"]
+
+			// then]
```
jakub-dzon, line 80: Please remove `]`

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
```
jakub-dzon, line 99: Please remove `]`

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
+				if env.ValueFrom.SecretKeyRef != nil {
+					return true, env.ValueFrom.SecretKeyRef.Optional, env.ValueFrom.SecretKeyRef.Name, env.ValueFrom.SecretKeyRef.Key
+				}
+				return false, nil, "", ""
+			})
+
+			// when
+			_, ok := maptypes["secret1"]
+
+			// then]
```
jakub-dzon, line 118: Please remove `]`

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
+				if env.ConfigMapRef != nil {
+					return true, env.ConfigMapRef.Optional, env.ConfigMapRef.Name
+				}
+				return false, nil, ""
+			})
+
+			// when
+			_, ok := maptypes["cm1"]
+
+			// then]
```
jakub-dzon, line 140: Please remove `]`

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
```
jakub-dzon, line 160: Please remove `]`

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
+					return true, env.SecretRef.Optional, env.SecretRef.Name
+				}
+				return false, nil, ""
+			})
+
+			// when
+			_, ok := maptypes["secret1"]
+
+			// then]
+			Expect(ok).To(BeTrue())
```
jakub-dzon, line 181: Please remove `]`

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
+				if volume.ConfigMap != nil {
+					return true, volume.ConfigMap.Optional, volume.ConfigMap.Name
+				}
+				return false, nil, ""
+			})
+
+			// when
+			_, ok := maptypes["cm1"]
+
+			// then]
```
jakub-dzon, line 202: Please remove `]`

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
```
jakub-dzon, line 222: Please remove `]`

internal/utils/utils_test.go
```diff
@@ -59,4 +60,187 @@ var _ = Describe("Utils", func() {
...
+				if volume.Secret != nil {
+					return true, volume.Secret.Optional, volume.Secret.SecretName
+				}
+				return false, nil, ""
+			})
+
+			// when
+			_, ok := maptypes["secret1"]
+
+			// then]
```
jakub-dzon, line 242: Please remove `]`

jakub-dzon (commented):
Thanks for the changes! 
Just a few more nitpicks

internal/configmaps/configmaps.go
jakub-dzon: That should be the same namespace as of the EdgeDeployment.

jakub-dzon (approved):
Looks good.
Please rebase.

machacekondra:
> Looks good. Please rebase.

done

pkliczewski (approved):

PR #85, opened 2022-01-26
Workload: Added AllowedList on metrics

eloycoto:
Each Workload can define where metrics are located, to make sure that
low bandwith is in use, users are allowed to filter the metrics and only
push the ones needeed.

Related-to: ECOPROJECT-495


api/v1alpha1/edgedeployment_types.go
```diff
@@ -67,6 +67,9 @@ type ContainerMetricsConfiguration struct {
 	// +kubebuilder:default=60
 	Interval int32 `json:"interval,omitempty"`
 
+	// Specification of workload metrics to be collected
+	AllowList []string `json:"allow_list,omitempty"`
```
gciavarrini, line 71: Why kubebuilder markers are not necessary here? Isn't it necessary set a default or control if the items in the allow list are unique (for example)?

eloycoto: Hi, 


> Why kubebuilder markers are not necessary here? Isn't it necessary set a default or control if the items in the allow list 
No default is needed. 

> are unique (for example)?

YUP, just added.

eloycoto: Umm, I cannot set the unique: 

Got this error:
```
<*errors.StatusError | 0xc000d78320>: {
            ErrStatus: {
                TypeMeta: {Kind: "", APIVersion: ""},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "CustomResourceDefinition.apiextensions.k8s.io \"edgedeployments.management.k4e.io\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[metrics].properties[allow_list].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic",
                Reason: "Invalid",
                Details: {
                    Name: "edgedeployments.management.k4e.io",
                    Group: "apiextensions.k8s.io",
                    Kind: "CustomResourceDefinition",
                    UID: "",
                    Causes: [
                        {
                            Type: "FieldValueForbidden",
                            Message: "Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic",
                            Field: "spec.validation.openAPIV3Schema.properties[spec].properties[metrics].properties[allow_list].uniqueItems",
                        },
                    ],
                    RetryAfterSeconds: 0,
                },
                Code: 422,
            },
        }
        CustomResourceDefinition.apiextensions.k8s.io "edgedeployments.management.k4e.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[metrics].properties[allow_list].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic
    occurred
```

gciavarrini: Let's skip this!

gciavarrini (approved):

pkliczewski (approved):

machacekondra (approved):

eloycoto:
I've just had a chat with @jakub-dzon and the idea is to use configmap instead of workload to keep the same way as ACM. I'll push the changes in a while.

jakub-dzon (changes requested):
As per ACM request for compatibility - please implement allow-list storage in a ConfigMap, as it's been done for system metrics.

jakub-dzon (approved):

jakub-dzon (approved):

PR #91, opened 2022-02-03
Check the kind cluster is properly up

machacekondra:
Previously the kind wasn't properly initialized, and we didn't know.


.github/workflows/kind.yaml
```diff
@@ -11,12 +11,16 @@ jobs:
...
+        with:
+          version: "v0.11.1"
       - name: Testing on KinD
         run: |
           export GO111MODULE=on
           export PATH=$PATH:$(go env GOPATH)/bin
           kubectl cluster-info
           kubectl get pods -n kube-system
+          kubectl wait --for=condition=Ready pods --all -n kube-system
+          kubectl get pods -n kube-system
```
eloycoto, line 23: You should delete the get pods in the line 21, no?

machacekondra: Maybe, I guess we can.

machacekondra: Actually, maybe better to remove line 23. So we know, which pods wasn't ok, before.. ?

.github/workflows/kind.yaml
```diff
@@ -11,12 +11,15 @@ jobs:
...
       - uses: engineerd/setup-kind@v0.5.0
+        with:
+          version: "v0.11.1"
       - name: Testing on KinD
         run: |
           export GO111MODULE=on
           export PATH=$PATH:$(go env GOPATH)/bin
           kubectl cluster-info
           kubectl get pods -n kube-system
+          kubectl wait --for=condition=Ready pods --all -n kube-system
```
masayag, line 22: should this be reduced to a specific set of pods?
if there is a chance that a pod might be in Completed, this condition might not return true for it.
but perhaps it isn't relevant for a fresh kind, and then this condition is safe.

masayag: I somehow missed the namespace. pls ignore the last comment.

eloycoto (approved):

jakub-dzon (approved):

PR #93, opened 2022-02-03
Auth: Registration with Certicate signed request

eloycoto:
This commit adds new parameters to the "registration" directive, where a user
will send a CSR to the endpoint, the operator will ask CA provider to get it
signed and return the certificate to the user.

At the moment, CSR is not required, and the main reason is to not break the
clients right now, but a follow-up commit will fix that and it'll implement
Authz and the missing tests.


internal/yggdrasil/yggdrasil.go
```diff
@@ -304,22 +307,49 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
...
 		logger.V(1).Info("received registration info", "content", registrationInfo)
+
+		res := models.MessageResponse{
+			Directive: msg.Directive,
+			MessageID: msg.MessageID,
+		}
+		content := models.RegistrationResponse{}
+
+		_, err = h.deviceRepository.Read(ctx, deviceID, h.initialNamespace)
+		if err == nil {
```
jakub-dzon, line 326: Shouldn't it be `err != nil`? `err == nil` means that the device has already been registered and in the past we ignored such registration request. From now on is that a way to re-issue a certificate?

internal/yggdrasil/yggdrasil_test.go
```diff
@@ -56,26 +65,52 @@ var _ = Describe("Yggdrasil", func() {
...
 		configMap          *configmaps.MockConfigMap
 
 		errorNotFound = errors.NewNotFound(schema.GroupResource{Group: "", Resource: "notfound"}, "notfound")
 		boolTrue      = true
+
+		k8sClient client.Client
+		testEnv   *envtest.Environment
 	)
 
+	initKubeConfig := func() {
```
jakub-dzon, line 79: Please change the name of the test file to `yggdrasil_integration_test`

internal/yggdrasil/yggdrasil.go
eloycoto: Certificate  expired, and we need to renew it. device worker already started to get 401 status code

pkliczewski (commented):
Do you think it makes sense to add some tests?

eloycoto:
> Do you think it makes sense to add some tests?

What I added has full test coverture and I'll add integration test too :-)

masayag (approved):

PR #95, opened 2022-02-04
Add integration test

machacekondra:
Current design of the integration test is using kind cluster in container as k8s provider for flotta-operator. For the devices we use containerized yggdrasil service, where the devices are running in privileged(currently) docker container. The container ships also podman, and systemd, so the deployments works fine in the container. To register the device user must run:

```bash
$ docker run -d --privileged quay.io/machacekondra/flotta-worker
```

This will register the device by default to k8s running on local port 8888. This could be configured by mounting proper yggdrasil configuration to `/etc/yggdrasil/` directory. We can also mount `/etc/machine-id` to emulate multiple devices.


test/e2e/e2e_test.go
```diff
@@ -0,0 +1,41 @@
...
+	var err error
+
+	BeforeEach(func() {
+		device, err = NewEdgeDevice("edgedevice1")
+		Expect(err).To(BeNil())
+	})
+
+	AfterEach(func() {
+		deviceId, err := device.GetId()
+		Expect(err).To(BeNil())
```
eloycoto, line 20: Nitpick: On AfterEAch we shouldn't run assertions. If for any reason cannot get the DeviceId, it'll stop and it'll be not `unregistered` or `remove()`


It shouldn't happen, you should be able to always get the device in this case, but I'll take extra care in these situations :-)

eloycoto: And I remembered to set up something like this on Cilium, we do the checks on JustAfterEach, and now it's part of Ginkgo 2.0, maybe we should update the version. 

https://onsi.github.io/ginkgo/#separating-diagnostics-collection-and-teardown-justaftereach



My approach will be: 



- JustAfterEach: Check race-conditions, validate device is still in there. 
- AfterEach: Force removing all devices, etc..



Also, with the new version you can also do AfterAll() BeforeAll(), so it'll be cool for these cases. :-)

pkliczewski (commented):
We have quay org we can use for any images https://quay.io/organization/project-flotta

test/e2e/common_test.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	time.Sleep(5 * time.Second)
+
+	return nil
+}
+
+func newClient() (dynamic.Interface, error) {
+	config, err := clientcmd.BuildConfigFromFlags("", "/home/omachace/.kube/config")
```
jakub-dzon, line 114: Shouldn't the path be different?

Makefile
```diff
@@ -118,6 +118,9 @@ gosec: ## Run gosec locally
 ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 test: manifests generate fmt vet test-fast ## Run tests.
 
+integration:
```
jakub-dzon, line 121: WDYT about making it obvious that this target runs tests (for example `integration-test`)?

test/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM quay.io/podman/stable
+
+WORKDIR /project
+
+# Setup Yggdrasil & device-worker repo:
+RUN curl -k https://copr.fedorainfracloud.org/coprs/ondramachacek/yggdrasil/repo/fedora-34/ondramachacek-yggdrasil-fedora-34.repo -o /etc/yum.repos.d/yggd.repo
```
jakub-dzon, line 6: We should have a non-personal copr account - not necessarily before this PR is merged.

test/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN curl -k https://copr.fedorainfracloud.org/coprs/eloyocoto/k4e-test/repo/fedora-34/eloyocoto-k4e-test-fedora-34.repo -o /etc/yum.repos.d/k4e-agent.repo && \
```
jakub-dzon, line 7: We should use `flotta` instead of `k4e` in names that we use for created files. (for example `k4e-agent.repo` -> `flotta-agent.repo`).

test/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# Certificate reqs:
+RUN mkdir /etc/pki/consumer && \
+    openssl req -new -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -out cert.pem -keyout key.pem -subj "/C=EU/ST=No/L=State/O=D/CN=www.example.com" && \
+    mv cert.pem key.pem /etc/pki/consumer
+
+# Default yggdrasil configuration should be replaced by volume with proper config:
+RUN echo 'key-file = "/etc/pki/consumer/key.pem"' >> /etc/yggdrasil/config.toml && \
+    echo 'cert-file = "/etc/pki/consumer/cert.pem"' >> /etc/yggdrasil/config.toml && \
+    echo 'client-id-source = "machine-id"' >> /etc/yggdrasil/config.toml && \
+    echo 'http-server = "172.17.0.1:8888"' >> /etc/yggdrasil/config.toml && \
```
jakub-dzon, line 27: Will that IP be always correct?

test/e2e/e2e_test.go
jakub-dzon: Unfortunately we need to migrate all of the tests before we can use Ginkgo 2 here. Let's have it as a separate story, maybe?

jakub-dzon (changes requested):

test/Dockerfile
machacekondra: This is default for docker installations. User can't mount custom configuration to `/etc/yggrasil` if needed.

test/e2e/common_test.go
machacekondra: Right! thanks

test/e2e/common_test.go
```diff
@@ -0,0 +1,126 @@
+package e2e_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/docker/docker/api/types"
```
pkliczewski, line 9: is there similar api for podman?

machacekondra: Yes. The API we already use in device-worker: "github.com/containers/podman/v3/pkg/bindings/containers"
But we use docker here, because github workflow is using docker by default, and also because I had some issues executing systemd in podman, but works fine in docker.

jakub-dzon (approved):
Nicely done!

pkliczewski (approved):

.github/workflows/kind.yaml
```diff
@@ -17,12 +17,34 @@ jobs:
...
           export PATH=$PATH:$(go env GOPATH)/bin
+          # Print cluster info
           kubectl cluster-info
           kubectl get pods -n kube-system
           kubectl wait --for=condition=Ready pods --all -n kube-system
-          make generate
-          make install
-          timeout 1m make run || retVal=$?
+
+          # TODO: Move to Makefile - Cert manager
```
masayag, line 25: heads up: this will be added on a PR next week to support non-OCP clusters with required resources.
it is planned to be part of the `make deploy` when no TARGET is specified or TARGET=k8s

.github/workflows/kind.yaml
```diff
@@ -17,12 +17,34 @@ jobs:
...
+          # Deploy flotta operator
+          make build
+          IMG=flotta-operator:latest make docker-build
+          kind load docker-image flotta-operator:latest
+          timeout 1m make deploy IMG=flotta-operator || retVal=$?
           if [ $retVal -ne 124 ]; then
             echo "Make run failed"; exit 1
           fi
+          kubectl wait --timeout=120s --for=condition=Ready pods --all -n flotta
+          kubectl port-forward deploy/flotta-operator-controller-manager -n flotta --address 0.0.0.0 8888:8888 &
```
masayag, line 38: for minikube we enable the nginx ingress controller and ingress CR to expose the service.
we'd might need a different [way to install](https://kind.sigs.k8s.io/docs/user/ingress/#ingress-nginx) for kind 
```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
```
unless we prefer port-forwarding instead for testing.

Makefile
```diff
@@ -132,7 +140,7 @@ deploy: gen-manifests ## Deploy controller to the K8s cluster specified in ~/.ku
 	kubectl apply -f $(TMP_ODIR)/flotta-operator.yaml
 ifeq ($(TARGET), k8s)
 ifneq (,$(shell which minikube 2>/dev/null))
-	minikube addons enable ingress
+	minikube addons enable ingress || true
```
masayag, line 143: perhaps we should add another possible value for `TARGET` as `kind`?
next week a PR will add cert-manager for `TARGET=k8s` as part of `make deploy` and will restructure the kustomize hierarchy to "kustomization" per cluster-type. 
The last to be merged will have to address this :-)

masayag (commented):
Very nice PR!

masayag (approved):

Makefile
masayag: Well... @arielireni let's add another TARGET for `kind` since it requires a different way to enable or deploy resources than minikube

.github/workflows/kind.yaml
machacekondra: I will add support for kind ingress. The port-forwarding was just quick working solution.

PR #97, opened 2022-02-07
TLS: Store tls server cert on K8s secrets

eloycoto:
Using this way, on each time that operator start, will retrieve the TLS
certificate from the provider, in this case, secrets. In the near
future, can be cert-manager, Vault, or any KMS.

Fix: ECOPROJECT-431

### DIFF
--- a/internal/mtls/ca.go
+++ b/internal/mtls/ca.go
@@ -37,6 +37,7 @@ type CAProvider interface {
 	GetCACertificate() (*CertificateGroup, error)
 	CreateRegistrationCertificate(name string) (map[string][]byte, error)
 	SignCSR(CSRPem string, commonName string, expiration time.Time) ([]byte, error)
+	GetServerCertificate(dnsNames []string, localhostEnabled bool) (*CertificateGroup, error)
 }
 
 type TLSConfig struct {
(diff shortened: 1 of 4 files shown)


internal/mtls/ca.go
```diff
@@ -37,6 +37,7 @@ type CAProvider interface {
 	GetCACertificate() (*CertificateGroup, error)
 	CreateRegistrationCertificate(name string) (map[string][]byte, error)
 	SignCSR(CSRPem string, commonName string, expiration time.Time) ([]byte, error)
+	GetServercertificate(dnsNames []string, localhostEnabled bool) (*CertificateGroup, error)
```
jakub-dzon, line 40: `GetServiceCertificate`

jakub-dzon (changes requested):
Just a little naming nit-picking.:)

internal/mtls/ca.go
masayag: @jakub-dzon  did you mean `GetServerCertificate` ? or to change the Server to Service?

jakub-dzon: Of course, `GetServerCertificate`. :facepalm:

jakub-dzon (approved):

pkliczewski (approved):

PR #98, opened 2022-02-09
TLS: Support ecdsa keys.

eloycoto:
By default, MTLS encryption only uses RSA keys. With this new way, it'll
support RSA and ECDSA keys, that has a better performance for the same
level of encryption.

Fix: ECOPROJECT-541

### DIFF
--- a/internal/mtls/ca_test.go
+++ b/internal/mtls/ca_test.go
@@ -1,7 +1,11 @@
 package mtls_test
 
 import (
+	"bytes"
 	"context"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/tls"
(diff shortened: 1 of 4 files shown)


pkliczewski:
Looks good to me. Let's make sure all the jobs are green.

masayag (approved):

jakub-dzon:
@eloycoto - please have a look at the failing test.

internal/mtls/ca_test.go
```diff
@@ -344,16 +348,44 @@ pNUUZ3cX4jGObCNfSBauHb73kfjGBo+RM2rnjAWqs9FhpnrJPDNT8Q==
 		Context("Sign CSR", func() {
 			var config *mtls.TLSConfig
 
-			BeforeEach(func() {
+			It("Sign valid pem with RSA keys", func() {
+				// Due to the CSR is created using RSA SignatureAlgorithm, we need to
```
jakub-dzon, line 352: nit: given/when/then

internal/mtls/ca_test.go
```diff
@@ -754,28 +800,27 @@ func createCACert() *certificate {
...
 	signedCert, err := x509.ParseCertificate(caBytes)
 	ExpectWithOffset(1, err).To(BeNil(), "Fail on parsing certificate")
 
-	return &certificate{ca, caPrivKey, caBytes, signedCert}
+	return &certificate{ca, key, caBytes, signedCert}
 }
 
 func createCSR() []byte {
-	keys, err := rsa.GenerateKey(rand.Reader, 1024)
+	// keys, err := rsa.GenerateKey(rand.Reader, 1024)
```
jakub-dzon, line 814: Please remove

PR #101, opened 2022-02-10
CI: Add coverage check

eloycoto:
On a new PR, a new check will be added with the percent of the test
coverage, it'll fail if the coverage is smaller than 80%.


jakub-dzon (approved):

machacekondra (approved):

jakub-dzon (approved):
**89.3** :smiling_face_with_tear:

PR #106, opened 2022-02-14
Add kind target

arielireni:
Add `TARGET=kind` and use it in `make deploy`.

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -26,11 +26,12 @@ jobs:
           make build
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
-          timeout 1m make deploy IMG=flotta-operator || retVal=$?
+          timeout 2m make deploy IMG=flotta-operator TARGET=kind || retVal=$?
           if [ $retVal -ne 124 ]; then
             echo "Make run failed"; exit 1
           fi
           kubectl wait --timeout=120s --for=condition=Ready pods --all -n flotta
+          # TODO: add ingress resource to config/kind/kustomization.yaml instead of port forwarding
           kubectl port-forward deploy/flotta-operator-controller-manager -n flotta --address 0.0.0.0 8888:8888 &
 
           # Run test
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ IMAGE_TAG_BASE ?= project-flotta.io/flotta-operator
 IMG ?= controller:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
-# Cluster type - k8s/ocp
+# Cluster type - k8s/ocp/kind
 TARGET ?= k8s
 # Host name for ingress creation
 HOST ?= flotta-operator.srv
@@ -141,12 +141,10 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 
 deploy: gen-manifests ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.yaml
-	kubectl wait --for=condition=Ready pods --all -n cert-manager
+	kubectl wait --for=condition=Ready pods --all -n cert-manager --timeout=60s
 	kubectl apply -f $(TMP_ODIR)/flotta-operator.yaml
 ifeq ($(TARGET), k8s)
-ifneq (,$(shell which minikube 2>/dev/null))
-	minikube addons enable ingress || true
-endif
+	minikube addons enable ingress
 endif
 
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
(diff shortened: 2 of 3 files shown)


masayag:
@arielireni pls rebase. the underlying PR was merged.

config/k8s/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+bases:
```
masayag, line 1: based on the other PR, let not use `bases` and move the path to ../base under the resources.

masayag, line 1: let's include the namespace and prefix here as well.

config/k8s/network/ingress.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: ingress
-  namespace: system
+  namespace: flotta
```
masayag, line 5: pls restore to `system`

config/kind/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+bases:
```
masayag, line 1: based on the other PR, let not use bases and move the path to ../base under the resources.

masayag (changes requested):

config/kind/kustomization.yaml
arielireni: done

.github/workflows/kind.yaml
```diff
@@ -26,11 +26,12 @@ jobs:
           make build
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
-          timeout 1m make deploy IMG=flotta-operator || retVal=$?
+          timeout 1m make deploy IMG=flotta-operator TARGET=kind || retVal=$?
```
masayag, line 29: @machacekondra do you have any idea how come the created file operator.yaml cannot be found? 
```
...
kubectl apply -f /tmp/tmp.SMmh0PSDvt/flotta-operator.yaml
error: the path "/tmp/tmp.SMmh0PSDvt/flotta-operator.yaml" does not exist
make: *** [Makefile:145: deploy] Error 1
Make run failed
Error: Process completed with exit code 1.
```
from https://github.com/project-flotta/flotta-operator/runs/5189563062?check_suite_focus=true#step:4:985

arielireni: fixed

Makefile
```diff
@@ -141,19 +141,19 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 
 deploy: gen-manifests ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.7.1/cert-manager.yaml
-	kubectl wait --for=condition=Ready pods --all -n cert-manager
+	kubectl wait --for=condition=Ready pods --all -n cert-manager --timeout=60s
```
arielireni, line 144: After adding kind target, I received this error:
`make: *** [Makefile:144: deploy] Terminated
error: no matching resources found`
Adding the timeout here instead of the timeout in kind.yaml [here](https://github.com/project-flotta/flotta-operator/pull/106/files#diff-e9c24daafcd1e27b49ed8503aee69248c36d7fc1a47ecbafffe36ba946c61f27L29 fixed the problem. @masayag WDYT?

masayag (approved):

.github/workflows/kind.yaml
```diff
@@ -26,11 +26,12 @@ jobs:
           make build
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
-          timeout 1m make deploy IMG=flotta-operator || retVal=$?
+          make deploy IMG=flotta-operator TARGET=kind || retVal=$?
```
masayag, line 29: pls change the order to:
`IMG=flotta-operator TARGET=kind make deploy`

arielireni: done

.github/workflows/kind.yaml
```diff
@@ -26,11 +26,12 @@ jobs:
           make build
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
-          timeout 1m make deploy IMG=flotta-operator || retVal=$?
+          IMG=flotta-operator TARGET=kind make deploy || retVal=$?
```
masayag, line 29: @machacekondra do you think we need to have a timeout for this? in that case it will have to be 2m (the originally 1m + 1m for the cert-manager deployment from within the `make deploy`)

machacekondra: Yes, I think we should keep it, and lets just increase it, as you suggest.

.github/workflows/kind.yaml
```diff
@@ -26,11 +26,12 @@ jobs:
           make build
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
-          timeout 1m make deploy IMG=flotta-operator || retVal=$?
+          timeout 2m IMG=flotta-operator TARGET=kind make deploy || retVal=$?
```
masayag, line 29: change to:
```
timeout 2m make deploy IMG=flotta-operator TARGET=kind || retVal=$?
```

arielireni: done

.github/workflows/kind.yaml
```diff
@@ -26,11 +26,12 @@ jobs:
           make build
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
-          timeout 1m make deploy IMG=flotta-operator || retVal=$?
+          timeout 2m make deploy IMG=flotta-operator TARGET=kind || retVal=$?
```
masayag, line 29: the condition below fails in the logs:
```
if [ $retVal -ne 124 ]; then
            echo "Make run failed";
fi
```

produces: 
```
/home/runner/work/_temp/38528907-5d0f-42d0-ac53-c99b6d373dca.sh: line 13: [: -ne: unary operator expected
```

masayag: it doesn't cause the job to fail, so I'll merge it, since the problem was here before this change. I'll create a ticket to address it.

PR #109, opened 2022-02-23
Add container image build and push job

machacekondra:

### DIFF
--- a/.github/workflows/container.yaml
+++ b/.github/workflows/container.yaml
@@ -1,0 +1,22 @@
+name: "Image upload"
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  kind:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out proper version of sources
+        uses: actions/checkout@v1
+
+      - name: Login to quay
+        env:
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          echo "$QUAY_TOKEN" | docker login quay.io --username project-flotta+github --password-stdin
+
+      - name: Build & push the container image to quay
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/container.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+      - name: Login to quay
+        env:
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+        run: |
+          echo "$QUAY_TOKEN" | docker login quay.io --username project-flotta+github --password-stdin
+
+      - name: Build & push the container image to quay
+        run: |
+          IMG=quay.io/project-flotta/flotta-operator make docker-build
```
masayag, line 21: should we specify a tag for the image, e.g. `dev`  ? or other notation?

machacekondra: Not sure, I thought the latest, which is default is fine. Do you think we should use dev?

masayag: My thought was to make sure we consume the image produced by this job.

masayag: So I didn't understand the usage of the this job.
I thought the purpose of this job is to produce an image for testing only.
Publishing an image to be consumed by any user with 'latest' tag and not tagging specifically (e.g. stable) might lead to instability.
If we intend to push an image to be consume by users, we should somehow make sure that:
1. It is known which agent version needs to be used.
2. It is pushed after it was passed tests.
In order to make the user experience more stable. 
ATM we don't have any versioning, so the assumption of using latest of all sides (operator and agent) should be okay.
But I think we'll see issues with it in the future.

masayag (approved):

pkliczewski (approved):

jakub-dzon (approved):

PR #113, opened 2022-02-25
Feat: added log collection options

eloycoto:
This commit add a way that a edgedevice is able to have a LogCollection
property, where workloads can retrieve logs and pushed on that way.

The definition of this task, is in the following RFC:

https://docs.google.com/document/d/1_hWAmLxZWkivIPe_XcoR6xxomuHFTqw4aB_A30MYOUc/edit

Related-to: ECOPROJECT-423


api/v1alpha1/edgedevice_types.go
```diff
@@ -34,9 +34,34 @@ type EdgeDeviceSpec struct {
...
+	BufferSize int32 `json:"BufferSize,omitempty"`
+
+	// SyslogConfig is the type to define Syslog Kind transport
+	SyslogConfig *LogSyslogConfig `json:"syslogConfig,omitempty"`
+}
+
+type LogSyslogConfig struct {
+
+	// Address with port to connect to syslog server.
+	Address string `json:"address,omitempty"`
```
masayag, line 60: WDYT of replacing it with a configmap (and for credentials when needed a secret), so when that configuration changes, the admin have to patch a single CR instead of all of the devices?

api/v1alpha1/edgedeployment_types.go
```diff
@@ -33,6 +33,7 @@ type EdgeDeploymentSpec struct {
 	Data            *DataConfiguration             `json:"data,omitempty"`
 	ImageRegistries *ImageRegistriesConfiguration  `json:"imageRegistries,omitempty"`
 	Metrics         *ContainerMetricsConfiguration `json:"metrics,omitempty"`
+	LogCollection   string                         `json:"log_collection,omitempty"`
```
jakub-dzon, line 36: Please add comment/docs what that string is supposed to be and what is its relation to configs in the EdgeDevice.

swagger.yaml
```diff
@@ -171,6 +171,27 @@ definitions:
...
+      LogCollection:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/logs-collection-information'
+          description: Log collection information
+
+  logs-collection-information:
+    type: object
+    properties:
+      kind:
```
jakub-dzon, line 183: Is `kind` needed if it is also a key in a map?

swagger.yaml
```diff
@@ -171,6 +171,27 @@ definitions:
       os:
         $ref: '#/definitions/os-information'
         description: OS lifecycle information
+      LogCollection:
```
jakub-dzon, line 174: Please change to lower case.

jakub-dzon (changes requested):

swagger.yaml
eloycoto: yes, can be two logs: 

```
log_collection:
  syslog-dc-foo:
    BufferSize: 10
    kind: syslog
    syslogConfig:
      address: dc-foo:514
      proto: tcp
  syslog-dc-bar:
    BufferSize: 10
    kind: syslog
    syslogConfig:
      address: dc-bar:514
      proto: tcp
````

Comment (jakub-dzon) [reply swagger.yaml:183]: OK, so the `kind` is not something user picks - it needs to come from a pre-defined set of values? In that case, WDYT about providing proper constants in `edgedevice_types.go`?

Comment (masayag) [reply swagger.yaml:183]: If the structure of each log-collector is different (Based on each collector implementation), it can be represented as with
`oneOf` and each collector will have its own definition.
Then we don't need to specify `kind`

File: api/v1alpha1/edgedevice_types.go [hunk 5]
```diff
@@ -34,9 +34,34 @@ type EdgeDeviceSpec struct {
...
+
+	// +kubebuilder:default=tcp
+	// +kubebuilder:validation:Enum=tcp;udp
+	Proto string `json:"proto,omitempty"`
```
jakub-dzon, line 64: Why not `Protocol`?

api/v1alpha1/edgedevice_types.go
```diff
@@ -34,9 +34,34 @@ type EdgeDeviceSpec struct {
...
+
+	// Kind is the type of log collection to be used
+	// +kubebuilder:validation:Enum=syslog
+	Kind string `json:"kind,omitempty"`
+
+	// +kubebuilder:default=12
+	// +kubebuilder:validation:Minimum=1
+	BufferSize int32 `json:"BufferSize,omitempty"`
+
+	// SyslogConfig is the type to define Syslog Kind transport
```
jakub-dzon, line 53: I think this comment should describe what NameRef should point to (ConfigMap).

config/crd/bases/management.project-flotta.io_edgedevices.yaml
```diff
@@ -53,6 +53,30 @@ spec:
                     format: int64
                     type: integer
                 type: object
+              logCollection:
+                additionalProperties:
+                  properties:
+                    BufferSize:
```
jakub-dzon, line 59: Lower case, please.

internal/yggdrasil/yggdrasil.go
```diff
@@ -675,3 +682,41 @@ func extractSecretsFromEnv(env []corev1.EnvVar, secretMap secretMapType) {
...
+		return nil, nil
+	}
+
+	res := map[string]models.LogsCollectionInformation{}
+	for key, val := range edgeDevice.Spec.LogCollection {
+		logConfig := models.LogsCollectionInformation{
+			BufferSize: val.BufferSize,
+			Kind:       val.Kind,
+		}
+		if val.SyslogConfig != nil {
```
jakub-dzon, line 697: WDYT about extracting the body of that if to a separate method?

internal/yggdrasil/yggdrasil_integration_test.go
```diff
@@ -493,6 +493,150 @@ var _ = Describe("Yggdrasil", func() {
...
+				Expect(logConfig.SyslogConfig.Protocol).To(Equal("tcp"))
+
+				Expect(config.Workloads).To(HaveLen(1))
+				workload := config.Workloads[0]
+				Expect(workload.Name).To(Equal("workload1"))
+				Expect(workload.LogCollection).To(Equal("syslog"))
+				Expect(workload.ImageRegistries).To(BeNil())
+			})
+
+			It("CM with invalid proto", func() {
```
jakub-dzon, line 582: nit: `protocol`

pkliczewski (approved):

masayag (approved):

api/v1alpha1/edgedevice_types.go
```diff
@@ -34,9 +34,24 @@ type EdgeDeviceSpec struct {
 	// RequestTime is the time of device registration request
 	RequestTime *metav1.Time `json:"requestTime,omitempty"`
 
-	Heartbeat *HeartbeatConfiguration `json:"heartbeat,omitempty"`
-	Storage   *Storage                `json:"storage,omitempty"`
-	Metrics   *MetricsConfiguration   `json:"metrics,omitempty"`
+	Heartbeat     *HeartbeatConfiguration         `json:"heartbeat,omitempty"`
+	Storage       *Storage                        `json:"storage,omitempty"`
+	Metrics       *MetricsConfiguration           `json:"metrics,omitempty"`
+	LogCollection map[string]*LogCollectionConfig `json:"log_collection,omitempty"`
```
machacekondra, line 40: Should we be consistent with naming and have here json:"logCollection"

masayag: You're right. the convention is camelCase.

masayag: Created https://github.com/project-flotta/flotta-operator/pull/124

PR #115, opened 2022-02-28
Add AnsiblePlaybook to DeviceConfigurationMessage

gciavarrini:
This PR adds AnsiblePlaybook field to the DeviceConfigurationMessage struct.

This change is needed for project-flotta/flotta-device-worker#69


jakub-dzon (approved):
I understand this is only an interim stage to allow for device worker changes to be mergable and there would be further changes as a result of #145.

gciavarrini:
@jakub-dzon exactly!

pkliczewski (approved):

PR #116, opened 2022-03-01
Ignore EdgeDevice status-only changes

jakub-dzon:
This PR introduces filtering for the `EdgeDevice` controller - from now on only changes in `EdgeDevice.spec` will trigger reconciliation.

Label changes are handled by the `EdgeDeviceLabelsReconciler`.

As per `GenerationChangedPredicate` documentation:
>  This predicate will skip update events that have no change in the object's metadata.generation field.
>  The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object.
> This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.

### DIFF
--- a/api/v1alpha1/edgedeployment_webhook.go
+++ b/api/v1alpha1/edgedeployment_webhook.go
@@ -96,7 +96,7 @@ func (r *EdgeDeployment) validate() error {
 		}
 
 		if _, ok := containersNames[container.Name]; ok {
-			return fmt.Errorf("name collisions for containers within the same pod spec are not supported.\n" +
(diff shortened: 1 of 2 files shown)


jakub-dzon:
@ydayagi - I think it is OK to ignore EdgeDevice label changes in the `EdgeDeviceReconciler` - is that correct?

ydayagi:
> @ydayagi - I think it is OK to ignore EdgeDevice label changes in the `EdgeDeviceReconciler` - is that correct?

@jakub-dzon indeed. Great

ydayagi (approved):

machacekondra (approved):

PR #120, opened 2022-03-02
Do not validate deployments on deletion

masayag:


config/webhook/manifests.yaml
```diff
@@ -23,7 +23,6 @@ webhooks:
     operations:
     - CREATE
     - UPDATE
-    - DELETE
```
jakub-dzon, line 26: :+1:

jakub-dzon (approved):
Good catch!

pkliczewski (approved):

jakub-dzon:
I think https://github.com/project-flotta/flotta-operator/blob/a1278aeff30d0a991d1db9e0ab8cf3f88b1bfb9e/api/v1alpha1/edgedeployment_webhook.go#L42 needs to be updated as well.

PR #128, opened 2022-03-09
Test: e2e gather always the logs for kind

eloycoto:
In this case we always gather the logs from kind even if the test fails
to be able to triage it.

Related-to: https://issues.redhat.com/browse/ECOPROJECT-649

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -37,10 +37,15 @@ jobs:
           # Run test
           make integration-test
 
-          # Export logs
+      # collect logs always just in case the test fails
+      - name: Collect all logs
+        if: always()
+        run: |
           kind export logs dist
+
       - name: Archive logs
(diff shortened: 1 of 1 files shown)


.github/workflows/kind.yaml
```diff
@@ -37,10 +37,15 @@ jobs:
           # Run test
           make integration-test
 
-          # Export logs
+      # collect logs always just in case the test fails
+      - name: Collect all logs
+        if: always()
```
gciavarrini, line 42: I'm not sure about this, but shouldn't expressions use the form `${{ <expression> }}`? In other words is it ok to use `always()` instead of `${{ always() }}` ?

gciavarrini: uh thanks! :)

gciavarrini (approved):

PR #130, opened 2022-03-10
Update ginkgo to v2

arielireni:
Updated to v2 in order to use JustAfterEach and OrderedContext.
The PR includes also changes made after `go mod tidy` & `go mod vendor`.


gciavarrini (approved):

go.mod
```diff
@@ -23,7 +23,8 @@ require (
 	github.com/kube-object-storage/lib-bucket-provisioner v0.0.0-20210818162813-3eee31c01875
 	github.com/morikuni/aec v1.0.0 // indirect
 	github.com/onsi/ginkgo v1.16.5
```
jakub-dzon, line 25: Why do we keep v1 dependency?

controllers/suite_test.go
```diff
@@ -47,37 +45,42 @@ var testEnv *envtest.Environment
...
-	Expect(err).NotTo(HaveOccurred())
 
-	err = obv1.AddToScheme(scheme.Scheme)
-	Expect(err).NotTo(HaveOccurred())
-
-	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
-	Expect(err).NotTo(HaveOccurred())
-}, 60)
+	done := make(chan interface{})
+	go func() {
```
masayag, line 55: Isn't there a risk that this function will be executed only after the test already started?

masayag: Unless `Eventually` enforces either its completion or its timeout :-)

eloycoto:
For the Kind job, looks like that the issue is that Go is not installed with the correct version, so I'll add the following to the action:

 
```
diff --git a/.github/workflows/kind.yaml b/.github/workflows/kind.yaml
index 38c82f4..b5e35df 100644
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -10,6 +10,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+
       - uses: engineerd/setup-kind@v0.5.0
         with:
           version: "v0.11.1"
```

So you're making sure that it's 1.16 as code.

masayag (approved):

gciavarrini:
Good job @arielireni! :)

controllers/suite_test.go
arielireni: yes it is, as described [here](https://onsi.github.io/gomega/#eventually)

PR #132, opened 2022-03-10
Enable profiling for the operator

masayag:
In certain cases using a profiler can be handy.
By default, the profiler is disabled.


masayag:
seems gosec isn't happy with this one. cloding.

docs/development/profiling.md
```diff
@@ -0,0 +1,75 @@
+# Profiling Flotta Operator
```
pkliczewski, line 1: @masayag What do you think about only documenting how to do it?

masayag: added: https://github.com/project-flotta/flotta-operator/pull/133

PR #136, opened 2022-03-14
Move mtls package from internal to pkg

eloycoto:
Two commits with the description. 

The main use case is to be able to use some functions on the device and do not duplicate things at all.


main.go
```diff
@@ -303,7 +303,10 @@ func main() {
 					if r.TLS != nil {
 						authType := yggdrasilAPIHandler.GetAuthType(r)
 						if !mtls.VerifyRequest(r, authType, opts, CACertChain) {
-							w.WriteHeader(http.StatusUnauthorized)
+							// w.WriteHeader(http.StatusUnauthorized)
+							// return
```
gciavarrini, line 307: Are these comments necessary?

pkg/mtls/caprovider_secret.go
```diff
@@ -32,7 +32,7 @@ const (
 	clientCertCertKey   = "client.crt"
 
 	certOrganization       = "flotta-operator"
-	certRegisterCN         = "register"
+	CertRegisterCN         = "register"
```
gciavarrini, line 35: Why should this const be exported? Will it be used in the device?

pkg/mtls/cert_utils.go
```diff
@@ -106,9 +106,9 @@ func (c *CertificateGroup) CreatePem() error {
 	return nil
 }
 
-func (c *CertificateGroup) marshalKeyToPem() (*bytes.Buffer, error) {
+func (c *CertificateGroup) MarshalKeyToPem(privKey crypto.PrivateKey) (*bytes.Buffer, error) {
 	privKeyPEM := new(bytes.Buffer)
-	switch t := c.privKey.(type) {
+	switch t := privKey.(type) {
```
gciavarrini, line 111: Question to learn: what is the reason of this change?

pkg/mtls/cert_utils.go
```diff
@@ -163,6 +163,10 @@ func (c *CertificateGroup) GetCert() *x509.Certificate {
 	return c.cert
 }
 
+func (c *CertificateGroup) GetKey() crypto.PrivateKey {
```
gciavarrini, line 166: Why do not export the `privKey` fields instead?

pkg/mtls/caprovider_secret.go
eloycoto: yes

main.go
eloycoto: When ECOPROJECT task will be in there, I'll uncomment, but I do not want to break the CI too much on device-changes.

masayag: perhaps worth adding FIXME or TODO comment to make this obvious (and avoid unintentional clean-up PRs by others)

masayag (approved):

main.go
eloycoto: Is just in the next line

masayag: :-) How did I miss it...

pkliczewski (approved):

PR #137, opened 2022-03-15
Propagate deployment labels to workload

machacekondra:
This PR add labels propagation from the edgedeployment CR to the
workload swagger specification.

The labels will be used to create the podman pod and so the pod could be
customized by specifing labels in the edgedeployment CR.

Issue: https://issues.redhat.com/browse/ECOPROJECT-281

For example if user want to create a pod with containers that has auto-update policy, he would create the edgedeployment as follows:

```yaml
apiVersion: management.project-flotta.io/v1alpha1
kind: EdgeDeployment
metadata:
  name: nginx
  labels:
    podman/io.containers.autoupdate: registry
spec:
  type: pod
  pod:
    spec:
      containers:
        - name: nginx
          image: quay.io/bitnami/nginx:latest
```

This will create a pod with label `io.containers.autoupdate: registry` and also all containers in the pod with such a label.

### DIFF
--- a/api/v1alpha1/edgedeployment_webhook_test.go
+++ b/api/v1alpha1/edgedeployment_webhook_test.go
@@ -204,6 +204,5 @@ var _ = Describe("EdgeDeployment Webhook", func() {
 			Expect(err).Should(MatchError("name collisions for containers within the same pod spec are not supported.\n" +
 				"container name: 'container' has been reused"))
 		})
-
 	})
 })
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -8,6 +8,7 @@ import (
 	"github.com/project-flotta/flotta-operator/internal/configmaps"
 	"github.com/project-flotta/flotta-operator/internal/devicemetrics"
 	"github.com/project-flotta/flotta-operator/internal/heartbeat"
+	"github.com/project-flotta/flotta-operator/internal/labels"
 	"github.com/project-flotta/flotta-operator/pkg/mtls"
 
 	"net/http"
@@ -492,6 +493,7 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
 		workload := models.Workload{
 			Name:          deployment.Name,
 			Namespace:     deployment.Namespace,
+			Labels:        labels.GetPodmanLabels(deployment.Labels),
 			Specification: string(podSpec),
 			Data:          data,
 			LogCollection: spec.LogCollection,
--- a/docs/design/http_api_swagger.md
+++ b/docs/design/http_api_swagger.md
@@ -981,11 +981,12 @@ Status: Internal Server Error
 
 | Name | Type | Go type | Required | Default | Description | Example |
 |------|------|---------|:--------:| ------- |-------------|---------|
-| configmaps | [ConfigmapList](#configmap-list)| `ConfigmapList` |  | |  |  |
-| data | [DataConfiguration](#data-configuration)| `DataConfiguration` |  | |  |  |
-| imageRegistries | [ImageRegistries](#image-registries)| `ImageRegistries` |  | |  |  |
+| configmaps | [ConfigmapList](#configmap-list)| `ConfigmapList` |  | | List of configmaps used by the workload |  |
+| data | [DataConfiguration](#data-configuration)| `DataConfiguration` |  | | Configuration for data transfer |  |
+| imageRegistries | [ImageRegistries](#image-registries)| `ImageRegistries` |  | | Image registries configuration |  |
+| labels | map of string| `map[string]string` |  | | Workload labels | Workload labels |
 | log_collection | string| `string` |  | | Log collection target for this workload |  |
-| metrics | [Metrics](#metrics)| `Metrics` |  | |  |  |
+| metrics | [Metrics](#metrics)| `Metrics` |  | | Metrics endpoint configuration |  |
 | name | string| `string` |  | | Name of the workload |  |
 | namespace | string| `string` |  | | Namespace of the workload |  |
 | specification | string| `string` |  | |  |  |
--- a/docs/user-guide/autoupdate.md
+++ b/docs/user-guide/autoupdate.md
@@ -1,0 +1,35 @@
+# Podman auto-update
+
+Auto update containers according to their auto-update policy.
+
+Auto-update looks up containers with a specified `io.containers.autoupdate` label. This label is set by the `flotta-operator`.
+Flotta-operator looks up labels on `EdgeDeployments` CR prefixed by `podman/` string. If such label is found it's propageted to
+the podman pod and containers.
+
+If  the  label  is  present  and set to registry, Podman reaches out to the corresponding registry to check if the image has been updated. The label image is an alternative to registry maintained for
+backwards compatibility.  An image is considered updated if the digest in the local storage is different than the one of the remote image.  If an image must be  updated,  Podman  pulls  it  down  and
+restarts the systemd unit executing the container.
+
+If `ImageRegistries.AuthFileSecret` is defined, Podman reaches out to the corresponding authfile when pulling images.
+
+Edgedevice start the `podman-auto-update.timer` which is responsible for executing the `podman-auto-update.service`. This unit is triggered daily.
+
+To create a Edgedeployment with auto-update feature enabled define following label:
+```yaml
+apiVersion: management.project-flotta.io/v1alpha1
+kind: EdgeDeployment
+metadata:
+  name: nginx
+  labels:
+    podman/io.containers.autoupdate: registry
+spec:
+  deviceSelector:
+    matchLabels:
+      device: local
+  type: pod
+  pod:
+    spec:
+      containers:
+        - name: nginx
+          image: quay.io/bitnami/nginx:1.20
+```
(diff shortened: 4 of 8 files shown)


pkliczewski (approved):

masayag:
@machacekondra how are the labels passed from the edgedeployment to the workload?

machacekondra:
> @machacekondra how are the labels passed from the edgedeployment to the workload?

Here: https://github.com/project-flotta/flotta-operator/pull/137/files#diff-ef2e06de9376c57c9524d086563a9579009a3251daf879c752093a91b34e883fR451

masayag:
> > @machacekondra how are the labels passed from the edgedeployment to the workload?
> 

so the entire list of labels is passed from EdgeDeployment to the workload, without any filtering on its content.
also, no prefix to figure out if a special trait needs to be passed on.
while this api simplifies things, it also passes more information than needed to the workload.

I prefer to add a prefix to the label which indicates it is relevant for the workload only.

In addition, the auto-update feature has also the option to specify `auth-file` by label `io.containers.autoupdate.authfile`.
Is there anything we can do from operator side to verify the required information is provided?
Any need to mount a file? Check the completeness of the EdgeDeployment CR based on this?

machacekondra:
> so the entire list of labels is passed from EdgeDeployment to the workload, without any filtering on its content. also, no prefix to figure out if a special trait needs to be passed on. while this api simplifies things, it also passes more information than needed to the workload.
> 
> I prefer to add a prefix to the label which indicates it is relevant for the workload only.

No problem to add prefix. Which do you suggest? For example `io.containers.*`? I was thinking about exposing a parameter in Edgedeployment CR, like 
```yaml
auto-update:
   type: registry
```
but in the end I thought propagating the labels would be generic and will work also with potential other `label` features. So maybe adding the filter `io.containers.*` would make sense here, if you agree.

> 
> In addition, the auto-update feature has also the option to specify `auth-file` by label `io.containers.autoupdate.authfile`. Is there anything we can do from operator side to verify the required information is provided? Any need to mount a file? Check the completeness of the EdgeDeployment CR based on this?

Not sure if we should do any validation if authfile is correct. Should we do validation just for existance, but not for a real validity? Or both?

masayag:
> but in the end I thought propagating the labels would be generic and will work also with potential other `label` features. So maybe adding the filter `io.containers.*` would make sense here, if you agree.

I thought of `podman/` as the prefix for any label that will be directed passed on to the workload.
e.g. `podman/io.containers.autoupdate=registry`
or any other label that the user might want to specify.


> Not sure if we should do any validation if authfile is correct. Should we do validation just for existance, but not for a real validity? Or both?

I wanted to verify only its existence. How does podman behave if the label is specified and file is missing?
I'd expect it to fail. If this isn't the case, maybe a bug needs to be reported to podman.

masayag:
Since it is going to be part of the API, pls describe it in a doc under https://github.com/project-flotta/flotta-operator/tree/main/docs/user-guide
Not sure if this one should go into [deploying-workloads.md](https://github.com/project-flotta/flotta-operator/blob/main/docs/user-guide/deploying-workloads.md) or [image-registries-auth.md](https://github.com/project-flotta/flotta-operator/blob/main/docs/user-guide/image-registries-auth.md)

internal/yggdrasil/yggdrasil.go
```diff
@@ -508,6 +515,33 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
...
+		}
+	}
+	return false
+}
+
+// getPodmanLabels filter all the labels of the EdgeDeployment CR starting with prefix "podman/"
+func getPodmanLabels(deploymentLabels map[string]string) map[string]string {
+	labels := map[string]string{}
+	for key, value := range deploymentLabels {
+		if strings.HasPrefix(key, "podman/") {
```
masayag, line 538: just an idea, since we have flottalabels.go, do you think it should go into there as well?
probably can be useful for reusing it from the unit-tests.

internal/yggdrasil/yggdrasil.go
```diff
@@ -508,6 +515,33 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
...
 }
 
+// validatePodmanLabels validate the podman labels
+func validatePodmanLabels(podmanLabels map[string]string, deployment v1alpha1.EdgeDeployment) bool {
+	for key, value := range podmanLabels {
+		if strings.Contains(key, "io.containers.autoupdate.authfile") {
+			for _, ctr := range deployment.Spec.Pod.Spec.Containers {
+				for _, vm := range ctr.VolumeMounts {
+					if value == vm.MountPath {
+						return true
```
masayag, line 525: if there are 2 labels and an authfile was provided only for the first one, this function returns true while it should fail.

internal/yggdrasil/yggdrasil.go
```diff
@@ -446,9 +446,16 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
 			data = &models.DataConfiguration{Paths: paths}
 		}
 
+		podmanLabels := getPodmanLabels(deployment.Labels)
+		if !validatePodmanLabels(podmanLabels, deployment) {
+			msg := "No volume with authfile path was found, please check if you mount the auth file."
+			h.recorder.Event(device, corev1.EventTypeWarning, "Misconfiguration", msg)
```
masayag, line 452: Is there an option to control the frequency of this event to prevent it from flooding (every 15 seconds)

internal/yggdrasil/yggdrasil.go
```diff
@@ -446,9 +446,16 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
 			data = &models.DataConfiguration{Paths: paths}
 		}
 
+		podmanLabels := getPodmanLabels(deployment.Labels)
```
masayag, line 449: Can you add tests for this pls?

masayag (changes requested):

internal/yggdrasil/yggdrasil.go
machacekondra: A logic for this would need to be introduced, there is nothing bultin. But I am thinking that such an event is not important for the user in the device Events IMHO, so I think it's ok, to be removed.

internal/yggdrasil/yggdrasil.go
machacekondra: You mean for example:
```yaml
labels:
  podman/io.containers.autoupdate.authfile: /tmp/auth
  labelX: value1
  labelY: value2

containers:
  ...
  mount:
    path: /tmp/auth
```

In such case we return true. What use case do you have exactly in mind?

masayag: My bad :-)
I mistakenly assumed that more than one authfile can be provided.

masayag:
Overall, LGTM

internal/yggdrasil/yggdrasil.go
```diff
@@ -446,9 +447,16 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
 			data = &models.DataConfiguration{Paths: paths}
 		}
 
+		podmanLabels := labels.GetPodmanLabels(deployment.Labels)
+		if !labels.ValidatePodmanLabels(podmanLabels, deployment) {
```
masayag, line 451: 2 thoughts about this one:
1. We can move the validation from here to the validating webhook of edge deployment. In that case, we won't allow user to create/update incorrect edgedeployment.
2. We don't need to record anything (event/log) in this context or perform any checks here. Else, if the deployment isn't valid, we'd might better return an error or filter the problematic deployment.
But the problem of 2 can be saved if we go with 1.
WDYT?

api/v1alpha1/edgedeployment_webhook_test.go
```diff
@@ -169,6 +169,9 @@ var _ = Describe("EdgeDeployment Webhook", func() {
 					},
 				}
 			}),
+			table.Entry("auth file label invalid", func() {
```
masayag, line 172: pls add another test for the positive case in which the label  is valid (one with only podman/label and one with ppodman/io.containers.autoupdate.authfile that is mounted properly) (iirc, they were in the previous version of the PR)

masayag (commented):
LGTM

masayag (approved):

jakub-dzon (commented):
How does that work with https://github.com/project-flotta/flotta-operator/blob/e00a492ce456d2b68e157a2f0ce064e961d6c874/internal/yggdrasil/yggdrasil.go#L46 current support for custom authfile?

machacekondra:
> How does that work with
> 
> https://github.com/project-flotta/flotta-operator/blob/e00a492ce456d2b68e157a2f0ce064e961d6c874/internal/yggdrasil/yggdrasil.go#L465
> current support for custom authfile?

That's a good question. Both work independently. Now question is should it? I can't think of a reason, where user would like to have two different autfiles, one for normal images and second for updates images, but I could be wrong.

So in general we could reuse authfile from custom authfile for podman autoupdates. We would need to just create volume for user, currently he is responsible. And we need to document it well, because the behavior will be different from podman and we won't be consistent.

jakub-dzon:
I think we should stick to one way of passing authfiles to the edgedevices in all use cases. Preferrably one that would be easier for the user.

machacekondra:
> I think we should stick to one way of passing authfiles to the edgedevices in all use cases. Preferrably one that would be easier for the user.

Ok, I will rework it.

jakub-dzon (approved):

PR #138, opened 2022-03-15
MTLS: prepare integration test to use MTLS certs 

eloycoto:
Until Yggdrasil RPM got updated it'll be using HTTP, but I want to avoid breaking the CI If I can, so this PR added a way to use certs on Yggdrasil, and also the correct ones: 

Commits: 

- Makefile: add a way to dump the certs
- Doc: example doc about how communication works on MTLS connection. 
- Test: copy some certs to the container before Yggdrasil starts. 
- Kind: listen on HTTPS port too (It'll be not be used for now) 

Related to : ECOPROJECT-399


docs/design/design.md
```diff
@@ -46,17 +46,92 @@ Agent is running in two processes:
...
+
+```mermaid
+sequenceDiagram
+autonumber
+participant Agent as Device Agent
+participant YGGD as Yggdrasil
+participant API as Operator API
+
+
+rect rgb(191, 223, 255)
```
jakub-dzon, line 59: Please use more neutral colors; the diagrams are difficult to read.

machacekondra (approved):

jakub-dzon (approved):

PR #140, opened 2022-03-16
Pass deployment namespace as workload namespace

machacekondra:
This commit add a new parameter namespace to workload model type, which
define the namespace of the deployment and should be used to distinguish
the workload on device-worker side.

### DIFF
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -448,6 +448,7 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
 
 		workload := models.Workload{
 			Name:          deployment.Name,
+			Namespace:     deployment.Namespace,
 			Specification: string(podSpec),
 			Data:          data,
 			LogCollection: spec.LogCollection,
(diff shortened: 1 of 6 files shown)


internal/yggdrasil/yggdrasil.go
```diff
@@ -448,6 +448,7 @@ func (h *Handler) toWorkloadList(ctx context.Context, logger logr.Logger, deploy
 
 		workload := models.Workload{
 			Name:          deployment.Name,
+			Namespace:     deployment.Namespace,
```
eloycoto, line 451: I'm happy with the change, but I think that it'll be good to add at least one assert on the integration test :-)

eloycoto (changes requested):
and an open question, what happens if a workload is deployed in a different namespace of the device?

machacekondra:
> and an open question, what happens if a workload is deployed in a different namespace of the device?

Currently it's not deployed at all if I am not mistaken.

eloycoto (approved):

pkliczewski (approved):

masayag (approved):

PR #141, opened 2022-03-17
Doc: Added MTLS running guide

eloycoto:

### DIFF
--- a/docs/user-guide/running.md
+++ b/docs/user-guide/running.md
@@ -45,7 +45,19 @@ To force device re-registration remove EdgeDevice CR from the cluster and `/var/
    `IMG=<image repository and tag> make install deploy`
 3. Forward flotta-operator ports to allow the agent to communicate with it:
  
-   `kubectl port-forward service/flotta-operator-controller-manager -n flotta 8888 --address 0.0.0.0`
+   `kubectl port-forward service/flotta-operator-controller-manager -n flotta 8043 --address 0.0.0.0`
+
+4. Dump the registration cert using the makefile target:
+
+```
+make get-certs
+```
+
+5. For running with Yggdrasil, the certs should have the correct permissions
+```
+sudo chown root:root /tmp/*.pem
+```
+
 
 ### yggdrasil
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/running.md
```diff
@@ -68,7 +80,31 @@ Following step should be done after flotta-device-worker is installed on the edg
...
+  --key-file /tmp/key.pem \
+  --ca-root /tmp/ca.pem \
+  --socket-addr @yggd \
+  --server 127.0.0.1:8043
+`
+
+When running Yggdrasil, a new edgedevice should be added into the list
+
+```
+kubectl get edgedevices --all-namespaces
```
pkliczewski, line 98: Aren't we creating edgedevices in default namespace for now? Why not to use -n default?

docs/user-guide/running.md
```diff
@@ -68,7 +80,31 @@ Following step should be done after flotta-device-worker is installed on the edg
...
+```
+
+And the certificate should be renewed, with one that it's specific for that
+device.
+
+```
+openssl x509 -noout -in cert.pem --subject
```
pkliczewski, line 105: maybe we could have a make target for this?

docs/user-guide/running.md
```diff
@@ -68,7 +80,31 @@ Following step should be done after flotta-device-worker is installed on the edg
...
+`
+sudo yggd \
+  --protocol http \
+  --path-prefix api/flotta-management/v1 \
+  --client-id $(cat /etc/machine-id) \
+  --cert-file /tmp/cert.pem \
+  --key-file /tmp/key.pem \
+  --ca-root /tmp/ca.pem \
+  --socket-addr @yggd \
+  --server 127.0.0.1:8043
```
gciavarrini, line 92: should this be similar to the commands decribed here https://github.com/project-flotta/flotta-operator/blob/092392deb17c39476c2ee24fd534f2e4ae334afc/README.md#inside-the-cluster?

machacekondra: I think we now use fqdn instead of IP.

docs/user-guide/running.md
```diff
@@ -47,6 +47,18 @@ To force device re-registration remove EdgeDevice CR from the cluster and `/var/
  
    `kubectl port-forward service/flotta-operator-controller-manager -n flotta 8888 --address 0.0.0.0`
```
machacekondra, line 48: s/8888/8043

docs/user-guide/running.md
eloycoto: We are using 8043 port forward on step 3, I do not think so.

eloycoto: yeah, I will

jakub-dzon (approved):

machacekondra (approved):

jakub-dzon (approved):

PR #143, opened 2022-03-17
e2e: Added network issues tests

eloycoto:
This commit adds two new 2e2 tests, that validates that an devices are
working on non ideal network environment.

The two cases added are the following ones:

- Connection has a delay of 200ms.
- Some packets drops from device to the operator API.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -49,7 +49,6 @@ var _ = Describe("e2e", func() {
 		_ = device.Unregister()
 		_ = device.Remove()
 	})
-
 	Context("Sanity", func() {
 		It("Deploy valid edgedeployment to registered device", func() {
 			// given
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -249,6 +248,44 @@ var _ = Describe("e2e", func() {
...
+			// Check the nginx is serving content:
+			stdout, err := device.Exec(fmt.Sprintf("curl http://localhost:%d", hostPort))
+			Expect(err).To(BeNil())
+			Expect(stdout).To(ContainSubstring("Welcome to nginx!"))
+		}
+
+		It("Packets delayed 200ms", func() {
+			// given
+			err := device.Register(
+				"dnf install iproute -y",
```
machacekondra, line 272: Do we want to add this to base test image?

machacekondra (approved):

test/e2e/e2e_test.go
eloycoto: should we add that for two test cases? I'm not sure about :sweat_smile:

bardielle (approved):

pkliczewski (approved):

test/e2e/e2e_test.go
```diff
@@ -249,6 +248,44 @@ var _ = Describe("e2e", func() {
...
+			Expect(err).To(BeNil())
+
+			conformanceTest()
+		})
+
+		It("Some packets drops", func() {
+			// given
+			err := device.Register(
+				"dnf install iproute -y",
+				"tc qdisc add dev eth0 root netem loss 10%")
```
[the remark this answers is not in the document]
eloycoto: The devices are just deleted when the test is finished. 
https://github.com/project-flotta/flotta-operator/blob/3c1cf3063a6826a3cc2e6de53e08ccec34ea86f9/test/e2e/e2e_test.go#L47-L49

masayag (approved):

test/e2e/e2e_test.go
machacekondra: Well, not really, but one day when we would wanna improve performance I think such things we will move ;)

masayag: When we'll start working on performance of the device-worker we might need additional packages.
In that case, it would be indeed faster to install all tests prerequisites during image creation time vs at test time.
Also, it from some reason, we'd like to perform the tests against ostree-based image, then this will fail.
But we can delay this to a future date when it will be needed.

PR #144, opened 2022-03-18
API: Authorization on API endpoints.

eloycoto:
When a device access the API, it can update ANY device. And that was a
serious problem.

This commit changes how authorization is done, so mtls.VerifyRequest
sets a new key on the context, so when a request access one resource
that does not match with the `cert.subject.CommonName` it'll be rejected
with a 403.

Fix: ECOPROJECT-402

### DIFF
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -363,6 +363,14 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 
 		_, err = h.deviceRepository.Read(ctx, deviceID, h.initialNamespace)
 		if err == nil {
+
+			if !IsOwnDevice(ctx, deviceID) {
+				// At this moment, the registration certificate it's no longer valid,
+				// because the CR is already created, and need to be a device
+				// certificate.
+				return operations.NewPostDataMessageForDeviceForbidden()
+			}
+
 			// @TODO remove this IF when MTLS is finished
 			if registrationInfo.CertificateRequest != "" {
 				cert, err := h.mtlsConfig.SignCSR(registrationInfo.CertificateRequest, deviceID)
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -11,7 +11,6 @@ import (
 	"github.com/project-flotta/flotta-operator/pkg/mtls"
 
 	"net/http"
-	"net/url"
 	"strings"
 
 	"time"
@@ -32,6 +31,7 @@ import (
 	"github.com/project-flotta/flotta-operator/internal/storage"
 	"github.com/project-flotta/flotta-operator/internal/utils"
 	"github.com/project-flotta/flotta-operator/models"
+	apioperations "github.com/project-flotta/flotta-operator/restapi/operations"
 	"github.com/project-flotta/flotta-operator/restapi/operations/yggdrasil"
 	operations "github.com/project-flotta/flotta-operator/restapi/operations/yggdrasil"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


pkg/mtls/ca.go
```diff
@@ -193,7 +194,7 @@ func isClientCertificateSigned(PeerCertificates []*x509.Certificate, CAChain []*
 // registration endpoint: Any cert signed, even if it's expired.
 // All endpoints: checking that it's valid certificate.
 // @TODO check here the list of rejected certificates.
-func VerifyRequest(r *http.Request, verifyType int, verifyOpts x509.VerifyOptions, CACertChain []*x509.Certificate) bool {
+func VerifyRequest(r *http.Request, verifyType int, verifyOpts x509.VerifyOptions, CACertChain []*x509.Certificate, authzKey string) bool {
```
pkliczewski, line 197: Do we plan to use any other `authzKey`? ATM we always use `yggdrasil.AuthzKey`

pkg/mtls/ca.go
```diff
@@ -204,14 +205,19 @@ func VerifyRequest(r *http.Request, verifyType int, verifyOpts x509.VerifyOption
...
 			return false
 		}
 		if _, err := cert.Verify(verifyOpts); err != nil {
 			// TODO log debug here with the error. Can be too verbose.
 			return false
 		}
+
+		if i == 0 && cert.Subject.CommonName != "" {
+			*r = *r.WithContext(context.WithValue(
+				r.Context(), authzKey, strings.ToLower(cert.Subject.CommonName)))
```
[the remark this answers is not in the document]
eloycoto: This is on context level. and the next one is the param of the URL. so to match always the same device.

internal/yggdrasil/yggdrasil.go
```diff
@@ -94,26 +96,45 @@ func NewYggdrasilHandler(deviceRepository edgedevice.Repository, deploymentRepos
...
+		if route.Operation.ID == YggrasilAPIRegistrationOperation {
+			return YggdrasilRegisterAuth
+		}
 	}
 	return res
 }
 
 func (h *Handler) GetControlMessageForDevice(ctx context.Context, params yggdrasil.GetControlMessageForDeviceParams) middleware.Responder {
 	deviceID := params.DeviceID
+	if !IsOwnDevice(ctx, deviceID) {
```
masayag, line 135: I'm debating with myself if an empty device-ID should be bad request or device-forbidden.
Can the device-id even be empty if it is part of the URL?

internal/yggdrasil/yggdrasil.go
```diff
@@ -141,6 +162,9 @@ func (h *Handler) GetControlMessageForDevice(ctx context.Context, params yggdras
 
 func (h *Handler) GetDataMessageForDevice(ctx context.Context, params yggdrasil.GetDataMessageForDeviceParams) middleware.Responder {
 	deviceID := params.DeviceID
+	if !IsOwnDevice(ctx, deviceID) {
```
masayag, line 165: Not sure if the generated code can allow it, but I don't like the fact we need to repeat the check in any call. Not that we plan on extending the APIs, but I'd prefer kind of a "template" function that authorization logic is located in a single place and all the calls have to go through it.
But not sure how the generated code can be customized to allow it.

internal/yggdrasil/yggdrasil.go
eloycoto: deviceID cannot be empty, is one validation on swagger already :-)

internal/yggdrasil/yggdrasil.go
eloycoto: We had an offline conversation here, it's hard to get the params.deviceID in a middleware call, and not in the autogenerated code too,  I cannot see a better way here :-( (I do not really liked too)

pkg/mtls/ca.go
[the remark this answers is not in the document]
pkliczewski: I understand a need. Why to do it here and in `IsOwnDevice`? We could have it only there?

pkg/mtls/ca.go
eloycoto: Just in case that we have a future RBAC with different key, owner for example.

pkg/mtls/ca.go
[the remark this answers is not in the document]
eloycoto: Here is to retrieve the DeviceID from certificate, and from URL is given by user. So. 

CERT: CN=foodevice path /XXX/foodevice/in: Allow. 
CERT: CN=bardevice path /XXX/bardevice/in: Block in IsOwnDevice()

jakub-dzon (approved):

internal/yggdrasil/yggdrasil.go
```diff
@@ -32,6 +31,7 @@ import (
 	"github.com/project-flotta/flotta-operator/internal/storage"
 	"github.com/project-flotta/flotta-operator/internal/utils"
 	"github.com/project-flotta/flotta-operator/models"
+	APIoperations "github.com/project-flotta/flotta-operator/restapi/operations"
```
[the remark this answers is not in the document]
masayag: @eloycoto ^^, perhaps even just `operations`

"Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns, such as:..."

https://go.dev/blog/package-names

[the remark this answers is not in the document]
eloycoto: operations is already used, this is why I changed.

masayag (approved):

PR #145, opened 2022-03-22
EdgePlaybook and PlaybookExecution CRDs definition

gciavarrini:
This PR aims to create all the necessary CRDs to execute custom playbooks on Flotta agents.


api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	Content string `json:"content"`
+	// The connection timeout on ansible-playbook
+	Timeout uint64 `json:"timeout,omitempty"`
+	// The required privelege level necessary to execute the playbook TODO: Enum like linux capabilities ?
+	RequiredPrivilegeLevel string `json:"requiredPrivilegeLevel,omitempty"`
+}
+
+// EdgePlaybookStatus defines the observed state of EdgePlaybook
+type EdgePlaybookStatus struct {
+	Phase                     string               `json:"phase,omitempty"`
```
pkliczewski, line 77: We may select many devices. How are we tracking status per device?

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	LastSyncedResourceVersion string               `json:"lastSyncedResourceVersion,omitempty"`
+	Hardware                  *Hardware            `json:"hardware,omitempty"`
+	Deployments               []PlaybookDeployment `json:"playbookDeployments,omitempty"`
+	DataOBC                   *string              `json:"dataObc,omitempty"`
+	UpgradeInformation        *UpgradeInformation  `json:"upgradeInformation,omitempty"`
+}
+
+type PlaybookDeployment struct {
+	Name               string              `json:"name"`
+	Phase              EdgeDeploymentPhase `json:"phase,omitempty"`
```
pkliczewski, line 88: What are phase values we want to support?

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+type AnsibleOptions struct {
+	// don't make any changes; instead, try to predict some of the changes that may occur
+	Check bool `json:"check,omitempty"`
+}
+
+type PrivilegeEscalationOptions struct {
+	Become bool `json:"become,omitempty"`
+	// Enum ksu, pbrun, enable, sesu, pmrun,runas, sudo, su, doas, pfexec, Machinectl, dzdo
```
pkliczewski, line 61: what values apply to rhel4edge or fedora iot?

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// Ansible binary file
+	BinaryPath string `json:"binaryPath,omitempty"`
+	// the ansible's playbooks list to be used
+	Playbooks []Playbook `json:"playbooks,omitempty"`
+	// the ansible's playbook options
+	Options *AnsibleOptions `json:"ansibleOptions,omitempty"`
+	// the ansible's playbook privilege escalation options
+	PrivilegeEscalationOptions *PrivilegeEscalationOptions `json:"privilegeEscalationOptions,omitempty"`
+	// enum that defines which is the stdout callback method (default = 'default'). Supported stdout methods are: debug,default,dense,json,minimal,null,oneline,stderr,timer,yaml,runner-events
+	StdoutCallback string `json:"stdoutCallback,omitempty"`
```
pkliczewski, line 51: We will collect output and send it back as events. Why do we need it?

gciavarrini: My doubt here is still: do we want to be compliant **only** with fleet management?

This field specifies the format of the events sent back. (Probably I choose a bad name for it 👎🏻  )

If we want to keep the possibility for the user to receive the "standard" output of `ansible-playbook`, then the user can specify the desidered `Stdoutcallback` (`json`,`yaml`, ...).
Specifying a `Stdoutcallback` equals to `runner-events` means that the user wants results compliant with fleet-management (which means translating from `ansible-playbook` output to `runner-events`). 

The information contained in the `ansible-playbook` output are much more than those in the `runner-events`, that's why I think it may be useful.

However I can remove the `Stdoutcallback` field and always sent the result as `runner-events`.
We can also decide to remote this field as first step and integrate other output format in the future.

I decided to leave this field in this PR so we would get a chance to discuss it :)

api/v1alpha1/edgeplaybook_types.go
gciavarrini: I was in doubt here: do we want to limit the available `BecomeMethod` or do we want to allow all the values and then reject the playbook during the validation process for a specif target environment?

api/v1alpha1/edgeplaybook_types.go
gciavarrini: Following @jakub-dzon 's advice:
>User would create CR defining the configuration of the Playbook, with matcher like for the EdgeDeployment and controller would create a number of PlaybookExecution CRs - one for device, tracking execution for that device.
Only thing is that in that case we make the scale problem worse.

I created a separate CRD `PlaybookExecution`.

I thought that a structure to indicate the status of `EdgePlaybook` CRD was necessary anyway.

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+type EdgePlaybookSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	DeviceSelector *metav1.LabelSelector `json:"deviceSelector,omitempty"`
+
+	// The ansible playbook command to execute
+	AnsiblePlaybookCmd *AnsiblePlaybookCmd `json:"ansiblePlaybookCmd,omitempty"`
+
+	LogCollection map[string]*LogCollectionConfig `json:"logCollection,omitempty"`
```
eloycoto, line 36: This is really needed?

api/v1alpha1/edgeplaybook_types.go
eloycoto: The thing is, stdout output to where?

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	BecomeUser   string `json:"becomeUser,omitempty"`
+}
+
+type Playbook struct {
+	// Playbook content
+	Content string `json:"content"`
+	// The connection timeout on ansible-playbook
+	Timeout uint64 `json:"timeout,omitempty"`
+	// The required privelege level necessary to execute the playbook TODO: Enum like linux capabilities ?
+	RequiredPrivilegeLevel string `json:"requiredPrivilegeLevel,omitempty"`
```
eloycoto, line 72: if this is linux capabilities should be an array of strings [CAP_ADMIN, CAP_NET_RAW, etc..]

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// enum that defines which is the stdout callback method (default = 'default'). Supported stdout methods are: debug,default,dense,json,minimal,null,oneline,stderr,timer,yaml,runner-events
+	StdoutCallback string `json:"stdoutCallback,omitempty"`
+}
+
+type AnsibleOptions struct {
+	// don't make any changes; instead, try to predict some of the changes that may occur
+	Check bool `json:"check,omitempty"`
+}
+
+type PrivilegeEscalationOptions struct {
```
eloycoto, line 59: what happens if the playbook has sudo:true in the yaml, can it scale the privileges?

api/v1alpha1/edgeplaybook_types.go
gciavarrini: Good question :D I'd like to have your opinion on it!
Of course I can remove it if you think it's not applicable/necessary

api/v1alpha1/edgeplaybook_types.go
gciavarrini: It can scale if the `User` privilege level is ` >=` of `RequiredPrivilegeLevel` . 
Keep in mind that there will be a validation step in the operator.

api/v1alpha1/edgeplaybook_types.go
pkliczewski: I am all for having support on the device but we can use one format and expose it either by events or as part of the status.

api/v1alpha1/edgeplaybook_types.go
pkliczewski: I would suggest to use values which makes sense in our context like sudo or su. Not sure if we need anything else. It could be supported by the device but we do not need to expose it on the CR.

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	Become bool `json:"become,omitempty"`
+	// Enum ksu, pbrun, enable, sesu, pmrun,runas, sudo, su, doas, pfexec, Machinectl, dzdo
+	BecomeMethod string `json:"becomeMethod,omitempty"`
+	BecomeUser   string `json:"becomeUser,omitempty"`
+}
+
+type Playbook struct {
+	// Playbook content
+	Content string `json:"content"`
```
machacekondra, line 68: I think this is not enough. Playbook could be simple one file, but could be also complex directory structure.

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// The ansible playbook command to execute
+	AnsiblePlaybookCmd *AnsiblePlaybookCmd `json:"ansiblePlaybookCmd,omitempty"`
+
+	LogCollection map[string]*LogCollectionConfig `json:"logCollection,omitempty"`
+}
+
+type AnsiblePlaybookCmd struct {
+	// username who execute the playbook
+	User string `json:"user,omitempty"`
```
machacekondra, line 41: What is difference between User nad BecomeUser in AnsibleOptions?

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	LastSeenTime              metav1.Time          `json:"lastSeenTime,omitempty"`
+	LastSyncedResourceVersion string               `json:"lastSyncedResourceVersion,omitempty"`
+	Hardware                  *Hardware            `json:"hardware,omitempty"`
+	Deployments               []PlaybookDeployment `json:"playbookDeployments,omitempty"`
+	DataOBC                   *string              `json:"dataObc,omitempty"`
+	UpgradeInformation        *UpgradeInformation  `json:"upgradeInformation,omitempty"`
+}
+
+type PlaybookDeployment struct {
+	Name               string              `json:"name"`
```
machacekondra, line 87: Should the user somehow define the name, so he knows which playbook CR to check for the status?

api/v1alpha1/edgeplaybook_types.go
gciavarrini: `User` is intended to be the user to connect to the host (can this apply to our context?)
`BecomeUser` allows to specify which user to use when executing a task with privilege escalation

api/v1alpha1/edgeplaybook_types.go
gciavarrini: I don't completely get the question. Can you clarify a bit? 

From an `EdgePlaybook` CR, the operator will generate several `PlaybookExecution` CR (one for each device on which the playbook must be run).

Where are you suggesting to put the field `Name`?

api/v1alpha1/edgeplaybook_types.go
gciavarrini: I'm sorry Piotr but I can completely understand you suggestion.
I get the idea, and I'm ok with it :) , but I can't understand how to implement that.

Can you give me an example, please?

In addition, ATM the Flotta agent always translate from json to runner-events. If we want to expose one format (i.e. `default`) and `runner-events` than we should plan to change that logic in the agent and to provide (in the operator?) a way to convert from `default` to all the other possibile format (i.e. `yaml`, `dense`, `timer`, `json`, ...). Or am i completely on the wrong track?

pkliczewski: As far as I understand this parameter it defines format of ansible output on the device. My point is that we will process the output on the device and send it back to operator so it can be exposed via kubeapi so I am not sure why user would want different formats.

Please correct me if I misunderstood something.

api/v1alpha1/edgeplaybook_types.go
machacekondra: Well, I am not sure how exactly local connection plugin works, so maybe it makes sense to have both. I know how that works for SSH, but unsure about local connection. So I think it's fine here.

api/v1alpha1/edgeplaybook_types.go
machacekondra: I think we could use following:
1) web link to a content I think archived, liked tar.gz file
2) Use the S3 as we do for workloads, but the other way around.
3) Use Ansible Tower/AWX links, I am not sure if they have such functionality, but maybe worth to explore.

I think for start we can use web link.

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	Phase              EdgeDeploymentPhase `json:"phase,omitempty"`
+	LastTransitionTime metav1.Time         `json:"lastTransitionTime,omitempty"`
+	LastDataUpload     metav1.Time         `json:"lastDataUpload,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// EdgePlaybook is the Schema for the EdgePlaybooks API
+type EdgePlaybook struct {
```
masayag, line 97: I just throwing an idea (don't want to cause too much noise to the PR), this name can be generalized to something such as `EdgeConfig` and now we support only configuration by ansible.

But if there will be another need for an additional method, this will remain fairly easy to extend.
It requires adding one level to the spec saying ansible, e.g.:
```
kind: EdgeConfig
...
spec:
  ansible:
    ... // EdgePlaybookSpec
  other_method: 
     ...
status:
   ...
```
we don't have any need for an alternative method for now, but we'd might want to have that option available (@rgolangh is exploring this)

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	AnsiblePlaybookCmd *AnsiblePlaybookCmd `json:"ansiblePlaybookCmd,omitempty"`
+
+	LogCollection map[string]*LogCollectionConfig `json:"logCollection,omitempty"`
+}
+
+type AnsiblePlaybookCmd struct {
+	// username who execute the playbook
+	User string `json:"user,omitempty"`
+	// Ansible binary file
+	BinaryPath string `json:"binaryPath,omitempty"`
```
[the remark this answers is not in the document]
gciavarrini: This can be useful both to specify the version to use and to specify a non standard ansible binary path.
We don't have customer requirements, so I'll try to figure out what the scenario may look like.

We can simply the CRD, at least in this step, by removing this field.
WDYT?

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// the ansible's playbooks list to be used
+	Playbooks []Playbook `json:"playbooks,omitempty"`
```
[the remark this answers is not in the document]
gciavarrini: Good point: thanks!
Yes, I think the execution order is important. 
The execution dependency is not present in this CRD. 

I'll try to fix these and I'll ask for you feedback again.

api/v1alpha1/edgeplaybook_types.go
eloycoto: But logs are going to be reported in Ansible CRD status in any way? Not sure about logs for execution TBH.

api/v1alpha1/edgeplaybook_types.go
eloycoto: I'm ok with it, but need to be` []string`

api/v1alpha1/edgeplaybook_types.go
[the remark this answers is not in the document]
jakub-dzon: IMO at this stage we should simplify as much as possible - to the bare minimum.

api/v1alpha1/edgeplaybook_types.go
jakub-dzon: I like that idea.

api/v1alpha1/edgeplaybook_types.go
jakub-dzon: Can we, for simplicity, start with in-line playbook and expand with options proposed by @machacekondra  later?

jakub-dzon (commented):
I am all for removing any knobs and switches that can have reasonable defaults in the device worker and pulling them to the CR only when there is an actual need.

api/v1alpha1/edgeplaybook_types.go
gciavarrini: My idea of relationship is: 
EdgeDevice * <-----> 1 EdgeExecution * <----> 1EdgePlaybook
@jakub-dzon do you have any opinion on this?

(regarding other question let me update the PR and then discuss about them)

api/v1alpha1/edgeplaybook_types.go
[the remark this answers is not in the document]
masayag: +1 for removing it.

api/v1alpha1/edgeplaybook_types.go
gciavarrini: I replaced this with a URL string (easy to undo) but in my opinion @jakub-dzon is right. We can make this CRD more complex later

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	RequiredPrivilegeLevel string `json:"requiredPrivilegeLevel,omitempty"`
+}
+
+// EdgePlaybookStatus defines the observed state of EdgePlaybook
+type EdgePlaybookStatus struct {
+	Phase                     string               `json:"phase,omitempty"`
+	LastSeenTime              metav1.Time          `json:"lastSeenTime,omitempty"`
+	LastSyncedResourceVersion string               `json:"lastSyncedResourceVersion,omitempty"`
+	Hardware                  *Hardware            `json:"hardware,omitempty"`
+	Deployments               []PlaybookDeployment `json:"playbookDeployments,omitempty"`
```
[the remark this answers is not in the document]
gciavarrini: None :)

api/v1alpha1/edgeplaybook_types.go
gciavarrini: @masayag I removed the `Phase` as suggested. Can you please check if it is correct?
Regarding the controller question I've added a "Open question" section to the gdocs (so I don't forget it :) )

api/v1alpha1/playbookexecution_types.go
```diff
@@ -25,19 +25,30 @@ import (
...
-	// Important: Run "make" to regenerate code after modifying this file
-
-	// Foo is an example field of PlaybookExecution. Edit playbookexecution_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	ExecutionStrategy ExecutionStrategy `json:"executionStrategy"`
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Default=0
+	ExecutionAttempt uint8 `json:"executionAttempt,omitempty" description:"the number of times the playbook has been executed" default:"0"`
+	// RunnerEvents     *models.AnsibleRunnerJobEventYaml `json:"runnerEvents,omitempty"`
+	// Events results.AnsiblePlaybookJSONResults `json:"events,omitempty"`
```
gciavarrini, line 33: I think this is the right place to put the fields related to the execution results: am I right?

`RunnerEvent` is the output compliant with fleet management. `RunnerEvents` type is generated in Makefile `generate-messages` target (from `internal/ansible/schema/ansibleRunnerJobEvent.yaml`).

`Events` are the results produced by `ansible-playbook` command.
`Events` type is `results.AnsiblePlaybookJSONResults` and it should be imported from `github.com/apenella/go-ansible/pkg/stdoutcallback/results`.

The problem is that, when I add `Events results.AnsiblePlaybookJSONResults` then the `make manifests` target fails as follow:

```bash
> make manifests                                                                                                                                                                                                                                                                    
gojsonschema --yaml-extension yaml -p models internal/ansible/schema/ansibleRunnerJobEvent.yaml -o internal/ansible/models/runner-job-event-gen.go 
/home/gloria/repos/flotta-operator/vendor/github.com/apenella/go-ansible/pkg/stdoutcallback/results/JSONResults.go:16:20: unsupported AST kind *ast.InterfaceType
/home/gloria/repos/flotta-operator/vendor/github.com/apenella/go-ansible/pkg/stdoutcallback/results/JSONResults.go:17:20: unsupported AST kind *ast.InterfaceType
/home/gloria/repos/flotta-operator/vendor/github.com/apenella/go-ansible/pkg/stdoutcallback/results/JSONResults.go:93:19: unsupported AST kind *ast.InterfaceType
/home/gloria/repos/flotta-operator/vendor/github.com/apenella/go-ansible/pkg/stdoutcallback/results/JSONResults.go:94:30: map values must be a named type, not *ast.InterfaceType
/home/gloria/repos/flotta-operator/vendor/github.com/apenella/go-ansible/pkg/stdoutcallback/results/JSONResults.go:99:19: unsupported AST kind *ast.InterfaceType
Error: not all generators ran successfully
run `controller-gen crd:trivialVersions=true,preserveUnknownFields=false rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases -w` to see all available markers, or `controller-gen crd:trivialVersions=true,preserveUnknownFields=false rbac:roleName=manager-role webhook paths=./... output:crd:artifacts:config=config/crd/bases -h` for usage
make: *** [Makefile:70: manifests] Error 1
```

The `unsupported AST kind *ast.InterfaceType` are related to `interface{}` fields. AFAIU, `controller-gen` can't handle `interface{}`. 
Do you have any suggestions?

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,159 @@
...
+	// Important: Run "make" to regenerate code after modifying this file.
+	EdgePlaybookStatus *EdgePlaybookStatus `json:"edgePlaybookStatus,omitempty"`
+}
+
+// EdgePlaybookSpec defines the desired state of EdgePlaybook
+type EdgePlaybookSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	DeviceSelector *metav1.LabelSelector `json:"deviceSelector,omitempty"`
```
masayag, line 47: IMO, with EdgeDeviceGroup the DeviceSelector should be part of EdgeDeviceGroup, or at the level of EdgeConfigSpec.
But to determine the proper place, we need to make sure both designs (for EdgeDeviceGroup and for EdgeConfig) are completing each other and playing well together.

api/v1alpha1/playbookexecution_types.go
masayag: Another question about `AnsibleRunnerJobEventYaml` - it has YAML in its name, but it is generated with json tags. Also, adding linter to verify the yaml can be useful.

Regarding the error with controller-gen, there is a discussion (linked inside) and solution for it here: https://github.com/kubernetes-sigs/controller-tools/issues/519#issuecomment-726489250
This will require to do few changes on go-ansible side.

gciavarrini: > Execution results can go into status or into events.

AFAIK, the status should not "store" data and could be generated on the fly at any moment. 
"The status summarizes the current state of the object in the system, and is usually persisted with the object by automated processes but may be generated on the fly. As a general guideline, fields in status should be the most recent observations of actual state, but they may contain information such as the results of allocations or similar operations which are executed in response to the object's spec."  ([API Conventions - Spec and Status](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status) )

So, if it's ok for you, I'll use Events.

> AnsibleRunnerJobEventYaml - it has YAML in its name

The name is auto generated and at the moment `go-jsonschema` doesn't allow customization. It basically transforms filename+extension  in its CamelCase (`ansibleRunnerJobEvent.yaml`  ->  `AnsibleRunnerJobEventYaml` )

> Also, adding linter to verify the yaml can be useful

Good idea! I'll add it.

> Regarding the error with controller-gen....

I think that using Events solves the problem since they don't have to be specified in CRD but in the `swagger.yaml` (correct?)

> This will require to do few changes on go-ansible side

Question to know your point of view: I saw that post but do you think an external library that legitimately uses the `interface{}` type can be changed to be `controller-gen` compliant? Is that a good enough motivation?

internal/ansible/schema/ansibleRunnerJobEvent.yaml
```diff
@@ -0,0 +1,98 @@
...
+        type: string
+      is_conditional:
+        type: boolean
+      uuid:
+        type: string
+        format: uuid
+      res:
+        type: object
+        properties:
+          _ansible_no_log:
```
masayag, line 72: should it start with underscore?

api/v1alpha1/playbookexecution_types.go
masayag: A library that 

> > Execution results can go into status or into events.
> 
> AFAIK, the status should not "store" data and could be generated on the fly at any moment. "The status summarizes the current state of the object in the system, and is usually persisted with the object by automated processes but may be generated on the fly. As a general guideline, fields in status should be the most recent observations of actual state, but they may contain information such as the results of allocations or similar operations which are executed in response to the object's spec." ([API Conventions - Spec and Status](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status) )

I was looking for this exact link :-) 

> 
> So, if it's ok for you, I'll use Events.
> 

+1

> > AnsibleRunnerJobEventYaml - it has YAML in its name
> 
> The name is auto generated and at the moment `go-jsonschema` doesn't allow customization. It basically transforms filename+extension in its CamelCase (`ansibleRunnerJobEvent.yaml` -> `AnsibleRunnerJobEventYaml` )
> 

thanks for clarifying 

> > Also, adding linter to verify the yaml can be useful
> 
> Good idea! I'll add it.
> 
> > Regarding the error with controller-gen....
> 
> I think that using Events solves the problem since they don't have to be specified in CRD but in the `swagger.yaml` (correct?)
> 

That should be correct. Events are managed by their own schema that we'll use.

> > This will require to do few changes on go-ansible side
> 
> Question to know your point of view: I saw that post but do you think an external library that legitimately uses the `interface{}` type can be changed to be `controller-gen` compliant? Is that a good enough motivation?

When we consume  library that doesn't satisfy all of our needs the options are as follow:
1. Open an issue that describe the problem and offer a PR to address it if agreed by the maintainers. 
2. Adapt our code.
3. Don't use it (seek for an alternative or develop own).
4. Fork it and adapt it.

The first option might lead to breaking the API of the library. If there is a good reasoning for doing it, then only in the next major version of that library the changes will be released.
But prior to it, reporting an issue with description of the problem can lead to the maintainer/member of the project to offer an alternative way or a remedy to the situation.

There need to be good reasoning for changing an API, but as I see it, the value of a library is to satisfy its consumers needs while making sense(this is too wide definition, but basically, not to damage dev principles :-)). A library without consumers is just a tool that will not be used.
If there is a majority of consumers for a library, than it will be hard to push a change that can affect multiple users.

The legitimacy will be derived from the consensus around that change.
We used to maintain libraries with only known 2 consumers other than us and it was easy to push changes and release frequently as requirements arrived.

And sometimes we'll find our self seeking for a very needed change requested by multiple users and the maintainers will not agree to. So the options are 2-4 which need to be well evaluated.

gciavarrini: Thanks @masayag!

internal/ansible/schema/ansibleRunnerJobEvent.yaml
gciavarrini: yes  :) 
[ansible doc](https://docs.ansible.com/ansible/2.5/dev_guide/developing_program_flow_modules.html#ansible-no-log)

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,151 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// EdgeConfigSpec defines the desired state of EdgeConfig
+type EdgeConfigSpec struct {
+	// The ansible playbook command to execute
+	EdgePlaybook *EdgePlaybookSpec `json:"edgePlaybook,omitempty"`
+
+	// EdgeDeviceGroup in which this EdgeConfig is applied
+	EdgeDeviceGroup *EdgeDeviceGroupSpec `json:"edgeDeviceGroup,omitempty"`
```
jakub-dzon, line 32: IMO that should be just a name of the group.

gciavarrini: Thanks! I'll change it!

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,151 @@
...
+type PrivilegeEscalationOptions struct {
+	Become bool `json:"become,omitempty"`
+	// +kubebuilder:validation:Enum=sudo;su
+	BecomeMethod string `json:"becomeMethod,omitempty"`
+	BecomeUser   string `json:"becomeUser,omitempty"`
+}
+
+type Playbook struct {
+	// Link to an arichived playbook content (tar.gz)
+	URL string `json:"url"`
```
gciavarrini, line 91: "Easy" plain text to start or URL to a `tar.gz`?
Personally, I prefer the plain text: we can always make things complex in other PRs but I'm not sure if this approach limits somehow future developments.
@jakub-dzon @machacekondra WDYT?

jakub-dzon: I'm all for the simple start - in-line plain text.

machacekondra: Sure, we can start with plaint text with limited functionality. I would just make sure we don't break the API in future, but I think having `URL` alongside with `Content` in `Playbook` type is OK.

api/v1alpha1/edgeplaybook_types.go
ydayagi: I think you can make the field '[]byte'. Think of it as an HTTP body. it can be anything. So it can be a single text file. It can be an archive of any type. this way in the future u can easily support the archives that the web link would have fetched. There are code libraries that will tell u the type of the content so u don't even need a separate field.

gciavarrini: Thanks @ydayagi for your suggestion! It make sense to me. 

Can you please give me an example of
>  "There are code libraries that will tell u the type of the content so u don't even need a separate field."

I've probably never used them :) 

@jakub-dzon @machacekondra do you agree with this?

ydayagi: well...the 'file' command in linux uses libmagic (or something like that). I'm sure there are libraries in GO. But even if u add a field for specifying the type it does not make a great difference. I googled and got this one: https://github.com/vimeo/go-magic

ydayagi: seems u can even use http.DetectContentType(bytes)

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,151 @@
...
+	// Playbook content
+	Content string `json:"content"`
+	// The connection timeout on ansible-playbook
+	Timeout uint64 `json:"timeout,omitempty"`
+	// TODO: Enum like linux capabilities ?
+	// The required privelege level necessary to execute the playbook
+	RequiredPrivilegeLevel []string `json:"requiredPrivilegeLevel,omitempty"`
+}
+
+type EdgePlaybookCondition struct {
```
jakub-dzon, line 99: I would love to have  documentation for possible condition state sets - not necessarily now, with the controller implementation at the latest.

api/v1alpha1/playbookexecution_types.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// PlaybookExecutionStatus defines the observed state of PlaybookExecution
+type PlaybookExecutionStatus struct {
+	Condition                 ExecutionPlaybookCondition `json:"condition,omitempty"`
```
jakub-dzon, line 38: That should be a slice of structs (like in the EdgeConfig). New conditions (of new type) should be appended to the slice and not replace the old one.
If singular state description is needed, that should be a `Phase`.

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// The ansible playbook command to execute
+	AnsiblePlaybookCmd *AnsiblePlaybookCmd `json:"ansiblePlaybookCmd,omitempty"`
+	//Execution strategy for each playbook
+	ExecutionStrategy map[string]ExecutionStrategy `json:"playbooksPriorityMap,omitempty"`
+}
+
+// EdgePlaybookStatus defines the observed state of EdgePlaybook
+type EdgePlaybookStatus struct {
+	Condition                 EdgePlaybookCondition `json:"condition,omitempty"`
```
jakub-dzon, line 53: That should be a slice of conditions as well.

api/v1alpha1/playbookexecution_types.go
```diff
@@ -0,0 +1,75 @@
...
+	LastSeenTime              metav1.Time                `json:"lastSeenTime,omitempty"`
+	LastSyncedResourceVersion string                     `json:"lastSyncedResourceVersion,omitempty"`
+	Completed                 bool                       `json:"completed,omitempty" description:"execution status, one of True, False"`
+}
+
+type ExecutionPlaybookCondition string
```
jakub-dzon, line 44: `PlaybookExecutionCondition`

api/v1alpha1/playbookexecution_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// RunnerEvents     *models.AnsibleRunnerJobEventYaml `json:"runnerEvents,omitempty"`
+	// Events results.AnsiblePlaybookJSONResults `json:"events,omitempty"`
+}
+
+// PlaybookExecutionStatus defines the observed state of PlaybookExecution
+type PlaybookExecutionStatus struct {
+	Condition                 ExecutionPlaybookCondition `json:"condition,omitempty"`
+	LastSeenTime              metav1.Time                `json:"lastSeenTime,omitempty"`
+	LastSyncedResourceVersion string                     `json:"lastSyncedResourceVersion,omitempty"`
+	Completed                 bool                       `json:"completed,omitempty" description:"execution status, one of True, False"`
```
jakub-dzon, line 41: That should probably be on of the condition types.

api/v1alpha1/playbookexecution_types.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type ExecutionPlaybookCondition string
+
+const (
+	ExecutionPlaybookDeploying EdgePlaybookConditionType = "ExecutionPlaybookDeploying"
```
jakub-dzon, line 47: These most likely should become condition types.

jakub-dzon (changes requested):

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,151 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// EdgeConfigSpec defines the desired state of EdgeConfig
+type EdgeConfigSpec struct {
+	// The ansible playbook command to execute
+	EdgePlaybook *EdgePlaybookSpec `json:"edgePlaybook,omitempty"`
+
+	// EdgeDeviceGroupName in which this EdgeConfig is applied
+	EdgeDeviceGroupName string `json:"edgeDeviceGroupName,omitempty"`
```
masayag, line 32: With a single group name where won't be an option to support multiple groups by the same config resource.

To support multiple groups by the same config, we can go with [same](https://github.com/project-flotta/flotta-operator/pull/180/files#diff-42d3e0337d69da396a4fc9def6162460e18795c57734394a89ec10a66c788630R80) solution for managing devices by a device group: Introducing a label `flotta/configured_by: edge_config_name` on all of the groups be managed by it.

Hence, from CRD pov, no attributed is needed. It will be addressed by the controller.

controllers/edgeconfig_controller.go
```diff
@@ -0,0 +1,62 @@
...
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the EdgeConfig object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
+func (r *EdgeConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
ydayagi, line 49: what is the purpose of this controller?

controllers/playbookexecution_controller.go
```diff
@@ -0,0 +1,62 @@
...
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the PlaybookExecution object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
+func (r *PlaybookExecutionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
ydayagi, line 49: what is the purpose of this controller?

config/crd/kustomization.yaml
```diff
@@ -13,13 +15,17 @@ patchesStrategicMerge:
...
+#- patches/webhook_in_edgeconfigs.yaml
+#- patches/webhook_in_playbookexecutions.yaml
 #+kubebuilder:scaffold:crdkustomizewebhookpatch
 
 # [CERTMANAGER] To enable webhook, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
 - patches/cainjection_in_edgedevices.yaml
 - patches/cainjection_in_edgedeployments.yaml
 #- patches/cainjection_in_edgedevicegroups.yaml
+#- patches/cainjection_in_edgeconfigs.yaml
```
eloycoto, line 27: these are really needed?

controllers/edgeconfig_controller.go
gciavarrini: I generated the controller, but I'll create another PR for this. 
Do you think It's better to remove it from this PR?

config/crd/kustomization.yaml
gciavarrini: Honestly.., I don't know. They were generated.
Do you think It's better to remove them and eventually add them in future PR (if needed)?

api/v1alpha1/edgeconfig_types.go
ydayagi: I don't know what is better. labels or specifying names in spec. However, I think it is better to do it the same as EdgeDevice and EdgeWorkload.

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,151 @@
...
+
+// EdgePlaybookSpec defines the desired state of EdgePlaybook
+type EdgePlaybookSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// The ansible playbook command to execute
+	AnsiblePlaybookCmd *AnsiblePlaybookCmd `json:"ansiblePlaybookCmd,omitempty"`
+	//Execution strategy for each playbook
+	ExecutionStrategy map[string]ExecutionStrategy `json:"playbooksPriorityMap,omitempty"`
```
ydayagi, line 48: I see we have multiple strategies. Don't we have just a single playbook here?

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,151 @@
...
+// EdgePlaybookStatus defines the observed state of EdgePlaybook
+type EdgePlaybookStatus struct {
+	Condition                 EdgePlaybookCondition `json:"condition,omitempty"`
+	LastSeenTime              metav1.Time           `json:"lastSeenTime,omitempty"`
+	LastSyncedResourceVersion string                `json:"lastSyncedResourceVersion,omitempty"`
+}
+
+type AnsiblePlaybookCmd struct {
+	// username who execute the playbook
+	User string `json:"user,omitempty"`
```
ydayagi, line 60: in EdgeWorkload we do not support specifying a user. What is the reason for doing it differently here? There's no mechanism within the playbook to specify user?

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,151 @@
...
+	// the ansible's playbooks list with execution priority
+	// +kubebuilder:validation:MinProperties=1
+	Playbooks map[string]Playbook `json:"playbooksPriorityMap,omitempty"`
```
ydayagi, line 63: so, we have multiple playbooks in a single playbookCmd? I think the strategy should be here as well.

config/crd/patches/webhook_in_edgeconfigs.yaml
```diff
@@ -0,0 +1,16 @@
+# The following patch enables a conversion webhook for the CRD
```
ydayagi, line 1: what is the reason for having a converting webhook?

config/crd/patches/webhook_in_playbookexecutions.yaml
```diff
@@ -0,0 +1,16 @@
+# The following patch enables a conversion webhook for the CRD
```
ydayagi, line 1: what is the reason for having a converting webhook?

ydayagi:
Perhaps it would be easier to push this as part of the initial implementation? I find it difficult to review without seeing the usage. Things change quickly when we start implementing. So maybe many things in this PR will automatically change during implementation.

api/v1alpha1/edgeconfig_types.go
```diff
@@ -93,6 +90,7 @@ type AnsibleOptions struct {
 }
 
 type RequiredPrivilegeLevel struct {
+	// See https://man7.org/linux/man-pages/man7/capabilities.7.html
 	CapAdd  []CapType `json:"capAdd,omitempty" description:"Capabilities to add"`
 	CapDrop []CapType `json:"capDrop,omitempty" description:"Capabilities to drop"`
```
gciavarrini, line 95: @eloycoto WDYT?

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,202 @@
...
+	LastSeenTime              metav1.Time             `json:"lastSeenTime,omitempty"`
+	LastSyncedResourceVersion string                  `json:"lastSyncedResourceVersion,omitempty"`
+}
+
+type AnsiblePlaybookCmd struct {
+	// username who execute the playbook
+	User string `json:"user,omitempty"`
+	// the ansible's playbooks list with execution priority
+	// +kubebuilder:validation:MinProperties=1
+	PlaybooksPriorityMap map[string]Playbook `json:"playbooksPriorityMap,omitempty"`
```
gciavarrini, line 60: @masayag regarding your previuous comment
> if we expect a list of playbooks, is there any importance for the order of execution or any dependency? (e.g. if 1/2 fails, should the second one be executed?)

I create a map in which the key is the priority of execution.
I'd like to use an integer as key but:
```
> make manifests
/home/gloria/repos/flotta-operator/api/v1alpha1/edgeconfig_types.go:60:27: map keys must be strings, not int
```

IS there a better way to handle order of execution?

api/v1alpha1/edgeplaybook_types.go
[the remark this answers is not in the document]
gciavarrini: See https://github.com/project-flotta/flotta-operator/pull/145/files#r847462381

api/v1alpha1/edgeplaybook_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// username who execute the playbook
+	User string `json:"user,omitempty"`
+	// Ansible binary file
+	BinaryPath string `json:"binaryPath,omitempty"`
+	// the ansible's playbooks list to be used
+	Playbooks []Playbook `json:"playbooks,omitempty"`
+	// the ansible's playbook options
+	Options *AnsibleOptions `json:"ansibleOptions,omitempty"`
+	// the ansible's playbook privilege escalation options
+	PrivilegeEscalationOptions *PrivilegeEscalationOptions `json:"privilegeEscalationOptions,omitempty"`
```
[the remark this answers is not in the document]
gciavarrini: Moved in `Playbook` struct

docs/design/crds.md
```diff
@@ -149,4 +149,76 @@ spec:
...
+      status: false
+      reason:
+      message: "Execution completed with error. Execution strategy: RetryOnFailure"
+      lastTransitionTime:
+    -
+      type: Completed
+      status: true
+      reason: "Execution completed with error. Execution strategy: ExecuteOnce"
+      message:
+      lastTransitionTime:
```
gciavarrini, line 223: @jakub-dzon do you have any suggestion on how to correctly define the statuses?

api/v1alpha1/edgeconfig_types.go
gciavarrini: See https://github.com/project-flotta/flotta-operator/pull/145/files#r847465346

api/v1alpha1/edgeconfig_types.go
gciavarrini: Done

api/v1alpha1/edgeconfig_types.go
gciavarrini: I removed the `EdgeConfigName` and I added the `DeviceIDs` field to manage device without a label group.
WDYT?

controllers/edgeconfig_controller.go
gciavarrini: Removed

controllers/playbookexecution_controller.go
gciavarrini: Removed

api/v1alpha1/edgeconfig_types.go
gciavarrini: You're right. I updated this section

api/v1alpha1/edgeconfig_types.go
gciavarrini: Moved `ExecutionStrategy` in Playbook struct

api/v1alpha1/edgeconfig_types.go
gciavarrini: We are executing custom playbooks so I think it can be useful to allow to specify a user to execute the playbook. 
What are the drawbacks of this in your opinion?

api/v1alpha1/edgeplaybook_types.go
gciavarrini: Done

api/v1alpha1/edgeconfig_types.go
ydayagi: I think that if u need an order then u should use a slice.

api/v1alpha1/edgeconfig_types.go
ydayagi: I have no opinion. I was just trying to understand how it works and perhaps make it easier for you in case a user is not required.

api/v1alpha1/edgeconfig_types.go
gciavarrini: @ydayagi yes it's a possibility! my concern is that, in that case, there is no "explicit" execution priority. 
Can be this a problem? WDYT?

docs/design/crds.md
jakub-dzon: There should be only one condition with any given type, with the most recent message and reason.

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,202 @@
...
+
+// EdgePlaybookSpec defines the desired state of EdgePlaybook
+type EdgePlaybookSpec struct {
+	// The ansible playbook command to execute
+	AnsiblePlaybookCmd *AnsiblePlaybookCmd `json:"ansiblePlaybookCmd,omitempty"`
+}
+
+// EdgePlaybookStatus defines the observed state of EdgePlaybook
+type EdgePlaybookStatus struct {
+	Condition                 []EdgePlaybookCondition `json:"condition,omitempty"`
```
jakub-dzon, line 50: `Conditions`, `json: conditions`

api/v1alpha1/playbookexecution_types.go
```diff
@@ -0,0 +1,87 @@
...
+type PlaybookExecutionSpec struct {
+	ExecutionStrategy ExecutionStrategy `json:"executionStrategy"`
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Default=0
+	ExecutionAttempt uint8 `json:"executionAttempt,omitempty" description:"the number of times the playbook has been executed" default:"0"`
+}
+
+// PlaybookExecutionStatus defines the observed state of PlaybookExecution
+type PlaybookExecutionStatus struct {
+	Condition                 []PlaybookExecutionCondition `json:"condition,omitempty"`
```
jakub-dzon, line 36: `Conditions`, `json: conditions`

pkliczewski:
@gciavarrini please rebase

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,185 @@
...
+
+// EdgeConfigSpec defines the desired state of EdgeConfig
+type EdgeConfigSpec struct {
+	// Definition of the custom playbook workload to execute
+	EdgePlaybook *EdgePlaybookSpec `json:"edgePlaybook,omitempty"`
+}
+
+// EdgeConfigStatus defines the observed state of EdgeConfig
+type EdgeConfigStatus struct {
+	EdgePlaybookStatus *EdgePlaybookStatus `json:"edgePlaybookStatus,omitempty"`
```
pkliczewski, line 34: What is the value of `EdgePlaybookStatus`?

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,185 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// EdgeConfigSpec defines the desired state of EdgeConfig
+type EdgeConfigSpec struct {
+	// Definition of the custom playbook workload to execute
+	EdgePlaybook *EdgePlaybookSpec `json:"edgePlaybook,omitempty"`
```
pkliczewski, line 29: What is the value of `EdgePlaybookSpec`?

pkliczewski (commented):
Overall looks good, 2 minor questions

api/v1alpha1/edgeconfig_types.go
gciavarrini: I'm not sure I understand your question/doubt correctly. 
If you're wondering why I created an `EdgePlaybookSpec` struct instead of putting the various fields directly into `EdgeConfigSpec` struct, it's because `EdgeConfig` is a generic CRD that can contain various types of configurations. At the moment we have only `EdgePlaybook` but in the future we could add more. So in this perspective, I preferred to create an `EdgePlaybookSpec` "sub-configuration". 
Did I answer your question? If not, could you help me to better understand what is your doubt?

pkliczewski: I understand and how we would model "various types of configurations"? Would we add more fields next to `EdgePlaybook`?

gciavarrini: Yes, that's what I had in mind. 
The structure will be 
```
type EdgeConfigSpec struct {
	EdgePlaybook *EdgePlaybookSpec 
        AnotherConfig  *AnotherConfigSpec
        ... etc ...
}
```

But if you think this isn't the best way to model additional configurations or if you think there might be drawbacks I'd be happy to get your opinion. 
PS. a similar argument applies to the `EdgePlaybookStatus`.

pkliczewski:
@jakub-dzon please take a look

docs/design/crds.md
```diff
@@ -149,4 +149,117 @@ spec:
...
+## EdgeConfig
+`EdgeConfig` is a namespaced custom resource that represents custom configuration that should be deployed to edge devices matching criteria specified in the CR.
+
+* apiVersion: `management.project-flotta.io/v1alpha1`
+* kind: `EdgeConfig`
+
+### Specification
+```yaml
+spec:
+  deviceIDs: # The deviceID list on which the playbook should be executed. Necessary to execute playbook on a devices that don't belong to any group
```
masayag, line 163: this seems outdated. no mention of `deviceIDs` in the PR.

main.go
```diff
@@ -29,6 +28,8 @@ import (
 
 	"github.com/kelseyhightower/envconfig"
 	routev1 "github.com/openshift/api/route/v1"
+	"go.uber.org/zap/zapcore"
```
masayag, line 31: Pls exclued this file from the PR.

masayag (changes requested):

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,185 @@
...
+	// +kubernetes:validation:Minimum=0
+	TimeoutSeconds int64 `json:"timeoutSeconds,omitempty"`
+	// The required privelege level necessary to execute the playbook
+	RequiredPrivilegeLevel *RequiredPrivilegeLevel `json:"requiredPrivilegeLevel,omitempty"`
+	// the ansible's playbook options for the playbook
+	Options *AnsibleOptions `json:"ansibleOptions,omitempty"`
+	// the ansible's playbook privilege escalation options for the playbook
+	PrivilegeEscalationOptions *PrivilegeEscalationOptions `json:"privilegeEscalationOptions,omitempty"`
+	//Execution strategy for the playbook
+	ExecutionStrategy ExecutionStrategy `json:"playbooksPriorityMap,omitempty"`
```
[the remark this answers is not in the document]
gciavarrini: Yes you're right. I'll fix them.

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,185 @@
...
+
+const (
+	StopAtFailuire ExecutionStrategy = "StopOnFailure"
+	RetryOnFailure ExecutionStrategy = "RetryOnFailure"
+	Once           ExecutionStrategy = "ExecuteOnce"
+)
+
+type AnsibleOptions struct {
+	// don't make any changes; instead, try to predict some of the changes that may occur
+	Check bool `json:"check,omitempty"`
```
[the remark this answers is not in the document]
gciavarrini: I agree that `DryRun` is better, but the ansible command is `Check` (https://docs.ansible.com/ansible/2.3/playbooks_checkmode.html). I thought to keep uniformity with ansible because it might prevent user confusion about what the option means. What are your thoughts on this?

api/v1alpha1/edgeconfig_types.go
```diff
@@ -0,0 +1,185 @@
...
+	Message *string `json:"message,omitempty" description:"human-readable message indicating details about last transition"`
+	// +optional
+	LastTransitionTime *metav1.Time `json:"lastTransitionTime,omitempty" description:"last time the condition transit from one status to another"`
+}
+
+type EdgePlaybookConditionType string
+
+const (
+	PlaybookDeploying EdgePlaybookConditionType = "Deploying"
+	Completed         EdgePlaybookConditionType = "Completed"
```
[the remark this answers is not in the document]
gciavarrini: No, here I'm modelling the fact that an EdgeConfig (with his EdgePlaybook) generate one or more PlaybookExecution.
`Completed` means that the EdgeConfig has been executed on all the involved devices.
If a playbook succeded or failed is modeled in the `PlaybookExecution` because it depends on the specific device  [see example](https://github.com/project-flotta/flotta-operator/blob/3f05bf9ef9c56d1e5b4630b2690a1aedf2934ccc/docs/design/crds.md#status-2).

In future PRs this can be improved (for example, the user can be interested in the percentage of deployments/completions of each EdgeConfig) but I'd like to start with a simple approach so I don't over-complicate this PR.

Makes sense for you?

docs/design/crds.md
gciavarrini: Done

main.go
gciavarrini: I reverted the changes. Is it ok now?

api/v1alpha1/edgeconfig_types.go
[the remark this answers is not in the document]
masayag: To make the decision harder, they mention both options (check and dry-run) in that link :-) 
If the convention of the CR to follow ansible options naming, then let's keep it as is.

eloycoto (approved):

jakub-dzon (approved):

masayag (approved):

pkliczewski (approved):

PR #149, opened 2022-03-24
Enhance flotta dashboard

masayag:

### DIFF
--- a/tools/deploy_grafana.sh
+++ b/tools/deploy_grafana.sh
@@ -102,8 +102,8 @@ spec:
      disable_signout_menu: true
    auth.anonymous: {}
    security:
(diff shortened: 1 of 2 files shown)


tools/deploy_grafana.sh
```diff
@@ -161,15 +162,15 @@ do
   fi
 done
 
+sleep 20
```
bardielle, line 165: Why do we need to wait here?

masayag: it takes some time for grafana handlers to be ready. I wasn't able to find an educated method to check when they are ready.
tried several methods. sleep seems to be the better one sadly.

bardielle (approved):

PR #150, opened 2022-03-25
Dashboard with Edge Device health (presence) metrics

jakub-dzon:
Follow-up for #146 adding Grafana dashboard panels for monitoring edge devices presence.

### DIFF
--- a/docs/metrics/flotta-devices-health.json
+++ b/docs/metrics/flotta-devices-health.json
@@ -1,0 +1,312 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": null,
+  "iteration": 1648459705287,
+  "links": [],
+  "panels": [
+    {
+      "datasource": null,
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "custom": {
+            "align": null,
+            "filterable": false
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 80
+              }
+            ]
+          }
+        },
+        "overrides": [
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "Value (lastNotNull)"
+            },
+            "properties": [
+              {
+                "id": "custom.displayMode",
+                "value": "color-text"
+              },
+              {
+                "id": "displayName",
+                "value": "Presence"
+              },
+              {
+                "id": "thresholds",
+                "value": {
+                  "mode": "absolute",
+                  "steps": [
+                    {
+                      "color": "dark-red",
+                      "value": null
+                    },
+                    {
+                      "color": "dark-green",
+                      "value": 1
+                    }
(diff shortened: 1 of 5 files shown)


masayag:
I wasn't able to open the dashboard in the browser with 40k devices.
Also, grafana's pod complains about:
```
t=2022-03-27T11:46:06+0000 lvl=eror msg="Request error" logger=context userId=1 orgId=1 uname=admin error="net/http: abort Handler"
t=2022-03-27T11:46:11+0000 lvl=eror msg="Request error" logger=context userId=1 orgId=1 uname=admin error="net/http: abort Handler"
t=2022-03-27T11:46:16+0000 lvl=eror msg="Request error" logger=context userId=1 orgId=1 uname=admin error="net/http: abort Handler"
```
The health panel will not report anything, while others do. 
Selection_1087

bardielle:
When I ran the deployment script it failed with the error- csv created in namespace with multiple operatorgroups, can't pick one automatically. I used that cmd- ```tools/deploy_grafana.sh docs/metrics/flotta-dashboard.json```

jakub-dzon:
> I wasn't able to open the dashboard in the browser with 23k devices. Also, grafana's pod complains about:

Warned you:)That's what I was expecting for large numbers of devices - panels for all devices are good for smaller numbers of them, but with larger numbers it's better to limit scope by filtering device IDs or namespaces. 
That said - it worked OK-ish for me with that number of devices when the time range was small (like 10,15, 30 minutes).

So - maybe a separate dashboard JSON with just these two panels plus a doc would be better?

jakub-dzon:
> When I ran the deployment script it failed with the error- csv created in namespace with multiple operatorgroups, can't pick one automatically. I used that cmd- `tools/deploy_grafana.sh docs/metrics/flotta-dashboard.json`

I think that's unrelated to the dashboard changes. @masayag - WDYT?

masayag:
> So - maybe a separate dashboard JSON with just these two panels plus a doc would be better?

It will be more useful for high scale. A parametrized dashboards that allows to select namespace will be more efficient in that case, e.g. how it is done in https://grafana.com/grafana/dashboards/11671
In our case namespace & time range can be configurable.

masayag:
> > When I ran the deployment script it failed with the error- csv created in namespace with multiple operatorgroups, can't pick one automatically. I used that cmd- `tools/deploy_grafana.sh docs/metrics/flotta-dashboard.json`
> 
> I think that's unrelated to the dashboard changes. @masayag - WDYT?

Agree. that's an issue with the script that was developed and tested against SNO only in which there were no operatorgroup under flotta namespace.

jakub-dzon:
@masayag - I've created a dedicated dashboard with filter box allowing users to pick desired deviceNamespace.

masayag (approved):

docs/metrics/flotta-devices-health.json
```diff
@@ -0,0 +1,312 @@
...
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "exemplar": true,
+          "expr": "ceil(sum by(deviceID, namespace)(rate(flotta_operator_edge_devices_heartbeat{job=~\"flotta.+\"}[2m])))",
```
masayag, line 237: should this be changed to `deviceNamespace` as well?

masayag (commented):
Removing ack since noticed the right panel doesn't preview data

docs/metrics/flotta-devices-health.json
masayag: This one with changing the query for the right pane:
Selection_1088

jakub-dzon: Yes!

jakub-dzon: Done

masayag (approved):

bardielle (approved):

PR #158, opened 2022-03-29
API: Log correctly

eloycoto:
Some kinds jobs were failing, and Yggdrasil logs has some weird 401 out
there. With this changes we're going to be able to debug what happens on
the operator side.

### DIFF
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -365,6 +365,8 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 		if err == nil {
 
 			if !IsOwnDevice(ctx, deviceID) {
+				authKeyVal, _ := ctx.Value(AuthzKey).(string)
+				logger.V(0).Info("Device tries to re-register with an invalid certificate", "certcn", authKeyVal)
 				// At this moment, the registration certificate it's no longer valid,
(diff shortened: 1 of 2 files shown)


internal/yggdrasil/yggdrasil.go
```diff
@@ -365,6 +365,8 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
 		if err == nil {
 
 			if !IsOwnDevice(ctx, deviceID) {
+				authKeyVal, _ := ctx.Value(AuthzKey).(string)
+				logger.V(0).Info("Device tries to re-register with an invalid certificate", "certcn", authKeyVal)
```
eloycoto, line 369: @masayag should I add metrics for this situation, registration with a renew-certificate, and an invalid one?

gciavarrini (approved):
lgtm

internal/yggdrasil/yggdrasil.go
masayag: That's a nice suggestion.
I'm thinking about expanding this not only to failed cert-renewal but for any violation of `IsOwnDevice` .
In that case, not only the registration API can be accessed using an invalid cert, but any other unauthorized access to the service.

eloycoto: :+1: I'll open a issue for that

PR #159, opened 2022-03-29
MTLS: Verify request errors

eloycoto:
When VerifyRequest failed there is no clear way to know if it was an invalid certificate, not present or something different.

These commits address a few things:

- A way to dump the secrets on GH actions.
- Fix the Verify function to validate why it fails. 
- A fix on mtls.CA to fix some flakes where CA certificate was not signed correctly.  
- Add the operator with debug level on GH actions.

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -23,6 +23,7 @@ jobs:
           kubectl wait --for=condition=Ready pods --all -n kube-system
 
           # Deploy flotta operator
+          sed -i 's/LOG_LEVEL=info/LOG_LEVEL=debug/g' config/manager/controller_manager_config.properties
           make build
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -43,6 +43,8 @@ jobs:
       - name: Collect all logs
         if: always()
         run: |
+          # Just dump the secrets to debug later
+          kubectl get secrets -n flotta -o json
           kind export logs dist
 
       - name: Archive logs
--- a/main.go
+++ b/main.go
@@ -316,8 +316,8 @@ func main() {
 				return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 					if r.TLS != nil {
 						authType := yggdrasilAPIHandler.GetAuthType(r, api)
-						if !mtls.VerifyRequest(r, authType, opts, CACertChain, yggdrasil.AuthzKey) {
-							setupLog.V(0).Info("Cannot verify request:", "authType", authType, "method", r.Method, "url", r.URL)
+						if ok, err := mtls.VerifyRequest(r, authType, opts, CACertChain, yggdrasil.AuthzKey); !ok {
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -316,8 +316,8 @@ func main() {
 				return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 					if r.TLS != nil {
 						authType := yggdrasilAPIHandler.GetAuthType(r, api)
-						if !mtls.VerifyRequest(r, authType, opts, CACertChain, yggdrasil.AuthzKey) {
-							setupLog.V(0).Info("Cannot verify request:", "authType", authType, "method", r.Method, "url", r.URL)
+						if ok, err := mtls.VerifyRequest(r, authType, opts, CACertChain, yggdrasil.AuthzKey); !ok {
+							setupLog.V(0).Info("Cannot verify request:", "authType", authType, "method", r.Method, "url", r.URL, "err", err)
```
gciavarrini, line 320: I see that `VerifyRequest` return an `err` only if `ok == false`. 
Do you think adding a comment on `VerifyRequest` can clarify why you aren't check the `err` value?

gciavarrini, line 320: s/Cannot/cannot/

machacekondra (approved):

.github/workflows/kind.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
           kubectl wait --for=condition=Ready pods --all -n kube-system
 
           # Deploy flotta operator
+          sed -i 's/LOG_LEVEL=info/LOG_LEVEL=debug/g' config/manager/controller_manager_config.properties
```
masayag, line 26: We can use this method to change log verbosity (without changing the codebase for it):
https://github.com/project-flotta/flotta-operator/blob/main/docs/user-guide/logger.md#change-the-verbosity-of-the-logger
```
kubectl patch cm -n flotta flotta-operator-manager-config --type merge --patch '{"data":{"LOG_LEVEL": "debug"}}'
```
it can be added after `make depoy` so the next command to wait for flotta pods will be there.

.github/workflows/kind.yaml
```diff
@@ -42,6 +43,8 @@ jobs:
       - name: Collect all logs
         if: always()
         run: |
+          # Just dump the secrets to debug later
+          kubectl get secrets -n flotta  -o json
```
masayag, line 47: sorry for the pickiness :-P but there are two spaces after `flotta` XD
Also, is json more convenient than yaml for debug? or just personal preference?

masayag (commented):
This will be very useful in debugging issues. Thank you!

.github/workflows/kind.yaml
gciavarrini: I had this doubt too, but I thought he wanted to avoid restarting the controller to speed up the kind task.

eloycoto: Gloria you're  right!

.github/workflows/kind.yaml
eloycoto: Json because is easy to write the certs to local using JQ

gciavarrini (approved):

masayag (approved):

Issue #153, opened 2022-03-27
Upgrade client-go and controller-runtime libraries


masayag:
In order to move to latest libraries, go version needs to be updated as well to 1.17.

- controller-runtime [changes](https://github.com/kubernetes-sigs/controller-runtime/releases/v0.11.0)

ahmadateya:
I have updated the go.mod file with 1.17 and removed the vendor directory to make sure it will use the new version.

running `go mod tidy` and `go mod vendor` updates the go.mod file with some new packages and edits other packages which mad some updates in the vendor directory, and running `make install` gives some warning about the deprecation of `go get` command but it works.

Now I tested the edits with `make install` and `make uninstall` commands but I am not sure if this upgrade may cause another issue because I am new to the flotta-operator and the project, So would you tell me how to test this upgrade?

this is my PR #172

PR #163, opened 2022-03-30
Upgrade go version to 1.17

arielireni:
This PR includes the following changes: 

- Update the go version to 1.17
- Replace `go get` with `go install`
- Update the version of the kustomise package in order to use `go install` instead of `go get`

The PR relates to #153 


masayag:
Few comments about this change:
1. Pls merge commits.
2. Go over https://tip.golang.org/doc/go1.17 and see recommendation that we'd like to apply (e.g. switch from `go get` to `go install`, run `gofmt` over the project, verify `go vet` and so...)

gabriel-farache (changes requested):
The [README](https://github.com/arielireni/flotta-operator/tree/upgrade_go_version#readme) shoud also be updated 
The file [go.yaml](https://github.com/arielireni/flotta-operator/blob/upgrade_go_version/.github/workflows/go.yml) is also referencing go-version and should be updated as well

jakub-dzon (approved):

arielireni:
> The [README](https://github.com/arielireni/flotta-operator/tree/upgrade_go_version#readme) shoud also be updated The file [go.yaml](https://github.com/arielireni/flotta-operator/blob/upgrade_go_version/.github/workflows/go.yml) is also referencing go-version and should be updated as well

Updated :)

config/crd/bases/management.project-flotta.io_edgeworkloads.yaml
```diff
@@ -472,11 +472,78 @@ spec:
                                                 ANDed.
                                               type: object
                                           type: object
+                                        namespaceSelector:
```
gabriel-farache, line 475: why did this file changed?

arielireni: This file is generated, I think the change here is because of the controller-gen upgrade

gabriel-farache (approved):

gciavarrini (approved):

gciavarrini:
Please rebase :)

config/crd/bases/management.project-flotta.io_edgeworkloads.yaml
arielireni: Yes :) this file changed only after running `make generate`, after upgrading `controller-gen`

PR #168, opened 2022-03-31
Add `JustAfterEach()` to e2e test

arielireni:
Added use of `JustAfterEach()` in order to indicate if there are race conditions reported to the log.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -40,15 +40,22 @@ var _ = Describe("e2e", func() {
 
 	})
 
-	AfterFailed(func() {
-		device.DumpLogs()
-	})
-
 	AfterEach(func() {
 		_ = deployment.RemoveAll()
 		_ = device.Unregister()
 		_ = device.Remove()
 	})
+
+	JustAfterEach(func() {
+		isValid, err := device.ValidateNoDataRaceInLogs()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(isValid).To(BeTrue(),"Found data race in logs" )
+	})
+
+	AfterFailed(func() {
+		device.DumpLogs()
+	})
+
 	Context("Sanity", func() {
 		It("Deploy valid edgedeployment to registered device", func() {
 			// given
--- a/test/e2e/edgedevice_test.go
+++ b/test/e2e/edgedevice_test.go
@@ -57,6 +57,7 @@ type EdgeDevice interface {
 	DumpLogs(extraCommands ...string)
 	Exec(string) (string, error)
 	WaitForDeploymentState(string, string) error
+	ValidateNoDataRaceInLogs(extraCommands ...string) (bool, error)
 }
 
 type edgeDeviceDocker struct {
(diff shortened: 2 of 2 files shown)


test/e2e/edgedevice_test.go
```diff
@@ -167,8 +170,19 @@ func (e *edgeDeviceDocker) DumpLogs(extraCommands ...string) {
...
+		logsMap[cmd] = output
+	}
+
+	return logsMap, err
+}
 
+func (e *edgeDeviceDocker) DumpLogs(extraCommands ...string) {
+	logsMap, err := e.GetLogs(extraCommands...)
+	if err != nil {
+		fmt.Printf("Error: GetLogs failed: %v \n", err)
```
masayag, line 182: Can we replace all `fmt.Printf` with `ginkgo.GinkgoT().Logd` ?

test/e2e/edgedevice_test.go
```diff
@@ -255,6 +269,20 @@ func (e *edgeDeviceDocker) Register(cmds ...string) error {
...
 }
 
+func (e *edgeDeviceDocker) ValidateNoDataRaceInLogs(extraCommands ...string) (bool, error) {
+	logsMap, err := e.GetLogs(extraCommands...)
+	if err != nil {
+		fmt.Printf("Error: GetLogs failed: %v \n", err)
+		return false, err
+	}
+	for _, output := range logsMap {
+		if strings.Contains(output, "WARNING: DATA RACE") {
```
masayag, line 279: will the race be logged?

test/e2e/edgedevice_test.go
arielireni: Done

test/e2e/edgedevice_test.go
eloycoto: shouldn't be needed at all TBH, because if test fail, it'll dump all logs :-)

test/e2e/e2e_test.go
```diff
@@ -49,6 +49,17 @@ var _ = Describe("e2e", func() {
 		_ = device.Unregister()
 		_ = device.Remove()
 	})
+
+	JustAfterEach(func() {
+		ret, err := device.ValidateNoDataRaceInLogs()
```
eloycoto, line 54: I would do: 
```
Expect(err).NotTo(HaveOccurred())
Expect(found).To(BeFalse(), "Found data race in logs")
```

eloycoto (commented):
Awesome work ariel :-)

test/e2e/e2e_test.go
arielireni: Done

masayag (approved):

jakub-dzon (approved):

test/e2e/e2e_test.go
```diff
@@ -49,6 +49,13 @@ var _ = Describe("e2e", func() {
 		_ = device.Unregister()
 		_ = device.Remove()
 	})
+
+	JustAfterEach(func() {
```
eloycoto, line 53: This change is great, but in this case, AfterFailed is not going to be executed, needs to be added *after* this `JustAfterEach()`

So, I would say that move AfterFailed function to the line 59, and if race happens, it'll be logged correctly :-)

PR #175, opened 2022-04-04
Rename `edgeDeployment` to `edgeWorkload`

arielireni:
Renamed `edgeDeployment` to `edgeWorkload` in order to avoid confusing with k8s deployment.


config/rbac/role.yaml
```diff
@@ -82,6 +82,12 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
```
machacekondra, line 85: I think you can remove this, it's already on line 36

masayag (approved):

machacekondra (approved):

docs/design/architecture.puml
```diff
@@ -1,13 +1,13 @@
 @startuml
-'https://plantuml.com/deployment-diagram
```
jakub-dzon, line 2: That should not be changed to `workload`

docs/osLifecycle.puml
```diff
@@ -1,5 +1,5 @@
 @startuml
-'https://plantuml.com/deployment-diagram
```
jakub-dzon, line 2: URL should remain unchanged.

jakub-dzon (changes requested):

jakub-dzon (approved):

PR #176, opened 2022-04-04
enable TLS for metrics remote writer

ydayagi:
pass an optional CA to the device
This is the CA that signs the metrics receiver's certificate

https://issues.redhat.com/browse/ECOPROJECT-579
enable TLS in device remote write

### DIFF
--- a/api/v1alpha1/edgedevice_types.go
+++ b/api/v1alpha1/edgedevice_types.go
@@ -44,6 +44,7 @@ type MetricsReceiverConfiguration struct {
 	RequestNumSamples int64  `json:"requestNumSamples,omitempty"`
 	TimeoutSeconds    int64  `json:"timeoutSeconds,omitempty"`
 	URL               string `json:"url,omitempty"`
+	CaSecretName      string `json:"caSecretName,omitempty"`
 }
 
 type LogCollectionConfig struct {
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 
-set -o nounset
 set -o pipefail
 set -o errexit
 
--- a/internal/yggdrasil/configuration-assembler.go
+++ b/internal/yggdrasil/configuration-assembler.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
+
 	"github.com/project-flotta/flotta-operator/internal/k8sclient"
 	"github.com/project-flotta/flotta-operator/internal/repository/edgedevicegroup"
 	"github.com/project-flotta/flotta-operator/internal/storage"
@@ -174,8 +176,14 @@ func (a *configurationAssembler) getDeviceMetricsConfiguration(ctx context.Conte
 	if deviceGroup != nil {
 		metricsConfigSpec = deviceGroup.Spec.Metrics
 	}
+
+	receiver, err := a.getMetricsReceiverConfiguration(ctx, metricsConfigSpec, edgeDevice.Namespace)
+	if err != nil {
+		return nil, err
+	}
+
 	metricsConfig := models.MetricsConfiguration{
-		Receiver: a.getMetricsReceiverConfiguration(metricsConfigSpec),
+		Receiver: receiver,
 	}
 
 	if metricsConfigSpec == nil {
@@ -491,7 +499,7 @@ func GetDefaultMetricsReceiver() *models.MetricsReceiverConfiguration {
 	}
 }
 
-func (a *configurationAssembler) getMetricsReceiverConfiguration(metrics *v1alpha1.MetricsConfiguration) *models.MetricsReceiverConfiguration {
+func (a *configurationAssembler) getMetricsReceiverConfiguration(ctx context.Context, metrics *v1alpha1.MetricsConfiguration, namespace string) (*models.MetricsReceiverConfiguration, error) {
 	result := GetDefaultMetricsReceiver()
 
 	if metrics != nil {
@@ -504,8 +512,26 @@ func (a *configurationAssembler) getMetricsReceiverConfiguration(metrics *v1alph
 				result.RequestNumSamples = receiverConfig.RequestNumSamples
 			}
 			result.URL = receiverConfig.URL
+
+			if result.URL != "" && strings.HasPrefix(result.URL, "https") && receiverConfig.CaSecretName != "" {
+				secret := corev1.Secret{ObjectMeta: metav1.ObjectMeta{
+					Name:      receiverConfig.CaSecretName,
+					Namespace: namespace,
+				}}
+				err := a.client.Get(ctx, client.ObjectKeyFromObject(&secret), &secret)
+				if err != nil {
+					return nil, err
+				}
+
+				caBytes := secret.Data["ca.crt"]
+				if len(caBytes) == 0 {
+					return nil, fmt.Errorf("metrics receiver config - missing key 'ca.crt' in secret %s", secret.Name)
+				}
+
+				result.CaCert = string(caBytes)
+			}
 		}
 	}
 
-	return result
+	return result, nil
 }
--- a/internal/yggdrasil/yggdrasil_integration_test.go
+++ b/internal/yggdrasil/yggdrasil_integration_test.go
@@ -9,10 +9,11 @@ import (
 	"crypto/x509/pkix"
 	"encoding/pem"
 	"fmt"
-	"github.com/project-flotta/flotta-operator/internal/repository/edgedevicegroup"
 	"path/filepath"
 	"strings"
 	"time"
+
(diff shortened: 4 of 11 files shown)


jakub-dzon:
@ydayagi - please rebase; that should resolve build failure

api/v1alpha1/edgedevice_types.go
```diff
@@ -44,6 +44,8 @@ type MetricsReceiverConfiguration struct {
 	RequestNumSamples int64  `json:"requestNumSamples,omitempty"`
 	TimeoutSeconds    int64  `json:"timeoutSeconds,omitempty"`
 	URL               string `json:"url,omitempty"`
+	// CA is a base64 encoding of CA certificate
+	CA string `json:"ca,omitempty"`
```
jakub-dzon, line 48: Please store the CA in a secret; this way, it could be easily shared and rotated more easily.

jakub-dzon (changes requested):

api/v1alpha1/edgedevice_types.go
ydayagi: I do not see a reason to store it in a secret. At least not at the moment. We do not know where the CA will come from nor what kind of receiver will be installed. In general, secrets are good just for one use case. Automatic updating of files in PODs/containers. In our case there's no pod. As far as I recall, applications such as certmanager can rotate configmaps just the same.
@pkliczewski @masayag @machacekondra would appreciate your opinion as well.

bardielle: There is already a CA cert store in a secret 
https://github.com/project-flotta/flotta-operator/blob/main/pkg/mtls/caprovider_secret.go#L93
Why need to store it in edgedevice?

eloycoto, line 48: Also, CA should be an array of strings, if it has an intermediate certificate need to have all the chain to get it working as expected.

jakub-dzon: I think it might be a different CA, used to create Thanos instance certificate and as such should be kept separately from device mtls CA.

ydayagi: @bardielle u raise an interesting point. @jakub-dzon I think you are right. However, in other features, e.g. data upload, we provided a default implementation and allowed the user to configure his own.
@bardielle point that Jakub raised regarding configuring EdgeDevice, was that user may want different metrics server for different devices/device-groups.

ydayagi: no need. the string can be a concatenation of CAs/certs. CA is saved to file and file can include multiple PEM certs/CAs.

pkliczewski: using secrets for CA seems to be common practice in k8s world

api/v1alpha1/edgedevice_types.go
```diff
@@ -44,6 +44,7 @@ type MetricsReceiverConfiguration struct {
 	RequestNumSamples int64  `json:"requestNumSamples,omitempty"`
 	TimeoutSeconds    int64  `json:"timeoutSeconds,omitempty"`
 	URL               string `json:"url,omitempty"`
+	SecretName        string `json:"secretName,omitempty"`
```
jakub-dzon, line 47: Are there plans to store any other certs in that secret besides CA? 
If not, please change the name to something more self-describing ('CASecretName`, `CASecret`, etc).

hack/generate.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -o nounset
+#set -o nounset
```
jakub-dzon, line 3: Please uncomment or remove the line.

api/v1alpha1/edgedevice_types.go
jakub-dzon, line 47: Please add it to the `EdgeDeviceGroup` as well.

jakub-dzon (changes requested):

api/v1alpha1/edgedevice_types.go
ydayagi: @jakub-dzon please provide some context. what is missing? where?

jakub-dzon: Sorry. My bad - I missed that the EdgeDeviceGroup crd file was updated as needed.

internal/yggdrasil/configuration-assembler.go
```diff
@@ -467,8 +473,26 @@ func (a *configurationAssembler) getMetricsReceiverConfiguration(device *v1alpha
 				result.RequestNumSamples = receiverConfig.RequestNumSamples
 			}
 			result.URL = receiverConfig.URL
+
+			if result.URL != "" && receiverConfig.CaSecretName != "" {
```
masayag, line 477: Do you think there should be a check that the scheme is https?

swagger.yaml
```diff
@@ -195,6 +195,8 @@ definitions:
         type: integer
       request_num_samples:
         type: integer
+      ca:
```
masayag, line 198: caCert will be a bit more clear name for this attribute, and if also CaSecretName --> CaCertSecretName

jakub-dzon (approved):

internal/yggdrasil/configuration-assembler.go
ydayagi: is it really that important? what if there's no scheme in URL? URL can be just host. That is if we want to be really strict and accurate.

swagger.yaml
ydayagi: come on.....CA is always a certificate. It is by definition a certificate. Why burden with so much text?

internal/yggdrasil/configuration-assembler.go
bardielle: I agree with @masayag

internal/yggdrasil/yggdrasil_integration_test.go
```diff
@@ -866,8 +866,49 @@ var _ = Describe("Yggdrasil", func() {
 					URL:               "http://metricsreceiver.io:19291/api/v1/receive",
 					RequestNumSamples: 1000,
 					TimeoutSeconds:    32,
+					CaSecretName:      "test",
+				}
+				expectedConfig := &models.MetricsReceiverConfiguration{
+					Ca:                "test",
```
bardielle, line 872: please move the variable "test" to a common variable (CaSecretNameTest)

internal/yggdrasil/yggdrasil_integration_test.go
```diff
@@ -866,8 +866,49 @@ var _ = Describe("Yggdrasil", func() {
...
+					RequestNumSamples: metricsReceiverConfig.RequestNumSamples,
+					TimeoutSeconds:    metricsReceiverConfig.TimeoutSeconds,
+					URL:               metricsReceiverConfig.URL,
+				}
+				device.Spec.Metrics = &v1alpha1.MetricsConfiguration{
+					ReceiverConfiguration: metricsReceiverConfig,
+				}
+				secretKey := client.ObjectKey{Namespace: testNamespace, Name: "test"}
```
bardielle, line 880: Same here -use a common var

internal/yggdrasil/yggdrasil_integration_test.go
```diff
@@ -866,8 +866,49 @@ var _ = Describe("Yggdrasil", func() {
...
+				Mockk8sClient.EXPECT().Get(gomock.Any(), secretKey, gomock.Any()).
+					Do(func(ctx context.Context, key client.ObjectKey, obj client.Object) {
+						Expect(obj).To(BeAssignableToTypeOf(&corev1.Secret{}))
+						secret := obj.(*corev1.Secret)
+						secret.Data = map[string][]byte{"ca.crt": []byte("test")}
+					}).
```
bardielle, line 886: test -> CaSecretKeyNameTest

internal/yggdrasil/yggdrasil_integration_test.go
```diff
@@ -882,6 +923,27 @@ var _ = Describe("Yggdrasil", func() {
...
 			})
 
+			It("receiver error reading secret", func() {
+				// given
+				device.Status.Deployments = nil
+				metricsReceiverConfig := &v1alpha1.MetricsReceiverConfiguration{
+					URL:               "http://metricsreceiver.io:19291/api/v1/receive",
+					RequestNumSamples: 1000,
+					TimeoutSeconds:    32,
+					CaSecretName:      "test",
```
bardielle, line 933: test-> common name as I mentioned before

swagger.yaml
bardielle: @ydayagi I agree with @masayag 
I think it's a terminology issue - you mentioned CA as a certificate but the terminology that we used in the code (as I know) is ca certificate

bardielle (approved):
Please fix the build job errors

PR #180, opened 2022-04-06
Edgedevice group configuration

jakub-dzon:
This PR introduces changes that allow for using `EdgeDeviceGroup` as a source of `EdgeDevice` configuration.


internal/yggdrasil/configuration-assembler.go
```diff
@@ -74,24 +76,32 @@ func (a *configurationAssembler) getDeviceConfiguration(ctx context.Context, edg
...
-
-	if edgeDevice.Spec.OsInformation != nil {
-		dc.Configuration.Os = (*models.OsInformation)(edgeDevice.Spec.OsInformation)
+	var deviceGroup *v1alpha1.EdgeDeviceGroup
+	if deviceGroupName, ok := edgeDevice.Labels["flotta/configured-by"]; ok {
+		logger.V(1).Info("Device uses EdgeDeviceGroup", "edgeDeviceGroup", deviceGroupName)
+		var err error
+		deviceGroup, err = a.groupRepository.Read(ctx, deviceGroupName, edgeDevice.Namespace)
+		if err != nil {
+			logger.Info("Cannot load EdgeDeviceGroup", "edgeDeviceGroup", deviceGroupName)
```
masayag, line 85: Shouldn't this be an error? after all it is a mis-configuration of the device CR.

internal/yggdrasil/configuration-assembler.go
```diff
@@ -74,24 +76,32 @@ func (a *configurationAssembler) getDeviceConfiguration(ctx context.Context, edg
...
+		}
 	}
 
-	err := a.setStorageConfiguration(ctx, edgeDevice, &dc)
+	dc.Configuration.Heartbeat = getHeartbeatConfiguration(edgeDevice, deviceGroup)
```
masayag, line 89: Pls document the behavior in the user guide, so the precedence of the configuration is known and its granularity

internal/yggdrasil/configuration-assembler.go
```diff
@@ -100,15 +110,32 @@ func (a *configurationAssembler) getDeviceConfiguration(ctx context.Context, edg
 	return &dc, nil
 }
 
-func getHeartbeatConfiguration(edgeDevice *v1alpha1.EdgeDevice) *models.HeartbeatConfiguration {
-	if edgeDevice.Spec.Heartbeat != nil {
+func getOsConfiguration(edgeDevice *v1alpha1.EdgeDevice, deviceGroup *v1alpha1.EdgeDeviceGroup) *models.OsInformation {
+	osInformation := edgeDevice.Spec.OsInformation
```
masayag, line 114: this means no patch-merge of group's configuration by device's configuration. 
for simplicity, we can start with it. but we might consider changing it in the future.

internal/yggdrasil/configuration-assembler.go
jakub-dzon: No. I took the eventual consistency approach - if it's missing it might get created at some point; for now serve configuration that is there.

internal/yggdrasil/configuration-assembler.go
jakub-dzon: Yes - that's a complete override. That was the intention - start small and modify when needed (supported by user feedback).

internal/yggdrasil/configuration-assembler.go
jakub-dzon: Done.

eloycoto (approved):

eloycoto (approved):

PR #181, opened 2022-04-08
Disable HTTP endpoint

eloycoto:
This commit disable all HTTP communication on flotta-operator API, and
enables only HTTPs connection using MTLS.

Fix: ECOPROJECT-646


ydayagi:
would be nice to have HTTP for development

masayag:
Also need to remove it from the deployment resources (https://github.com/project-flotta/flotta-operator/blob/main/config/manager/manager.yaml#L39) from podspec, service and route/ingress

ydayagi:
> Also need to remove it from the deployment resources (https://github.com/project-flotta/flotta-operator/blob/main/config/manager/manager.yaml#L39) from podspec, service and route/ingress

@masayag can't we have HTTP for development?

masayag:
> > Also need to remove it from the deployment resources (https://github.com/project-flotta/flotta-operator/blob/main/config/manager/manager.yaml#L39) from podspec, service and route/ingress
> 
> @masayag can't we have HTTP for development?

I don't think it can be supported, since 

> > Also need to remove it from the deployment resources (https://github.com/project-flotta/flotta-operator/blob/main/config/manager/manager.yaml#L39) from podspec, service and route/ingress
> 
> @masayag can't we have HTTP for development?

With the current code it isn't possible due to YggdrasilHandler at https://github.com/project-flotta/flotta-operator/blob/main/main.go#L322. At some point iinw, the certificate will contain a required piece of information for the approval of the device.
@ydayagi pls specify the pain points and let's see what can be done to ease them (there is some work around it already)

eloycoto:
No, HTTP interface is no longer a option.

ydayagi:
> > > Also need to remove it from the deployment resources (https://github.com/project-flotta/flotta-operator/blob/main/config/manager/manager.yaml#L39) from podspec, service and route/ingress
> > 
> > 
> > @masayag can't we have HTTP for development?
> 
> I don't think it can be supported, since
> 
> > > Also need to remove it from the deployment resources (https://github.com/project-flotta/flotta-operator/blob/main/config/manager/manager.yaml#L39) from podspec, service and route/ingress
> > 
> > 
> > @masayag can't we have HTTP for development?
> 
> With the current code it isn't possible due to YggdrasilHandler at https://github.com/project-flotta/flotta-operator/blob/main/main.go#L322. At some point iinw, the certificate will contain a required piece of information for the approval of the device. @ydayagi pls specify the pain points and let's see what can be done to ease them (there is some work around it already)

as a developer, I would like to execute 'make run' for the operator wherever I like. Then execute yggdrasil wherever I like, just by executing the executable file. Have any kind of cluster that is k8s compatible. 
At the moment I have to create certs, copy them to wherever yggdrasil is etc. I don't want to depend on our make targets.
Every make command I use fails now and then cause our team changes it rapidly and introduces all kind of dependencies that are only relevant to specific developers.
Now, if there's an error.....the logs will only throw you in irrelevant directions.

eloycoto:
> > > > Also need to remove it from the deployment resources (https://github.com/project-flotta/flotta-operator/blob/main/config/manager/manager.yaml#L39) from podspec, service and route/ingress
> > > 
> > > 
> > > @masayag can't we have HTTP for development?
> > 
> > 
> > I don't think it can be supported, since
> > > > Also need to remove it from the deployment resources (https://github.com/project-flotta/flotta-operator/blob/main/config/manager/manager.yaml#L39) from podspec, service and route/ingress
> > > 
> > > 
> > > @masayag can't we have HTTP for development?
> > 
> > 
> > With the current code it isn't possible due to YggdrasilHandler at https://github.com/project-flotta/flotta-operator/blob/main/main.go#L322. At some point iinw, the certificate will contain a required piece of information for the approval of the device. @ydayagi pls specify the pain points and let's see what can be done to ease them (there is some work around it already)
> 
> as a developer, I would like to execute 'make run' for the operator wherever I like. Then execute yggdrasil wherever I like, just by executing the executable file. Have any kind of cluster that is k8s compatible. At the moment I have to create certs, copy them to wherever yggdrasil is etc. I don't want to depend on our make targets. Every make command I use fails now and then cause our team changes it rapidly and introduces all kind of dependencies that are only relevant to specific developers. Now, if there's an error.....the logs will only throw you in irrelevant directions.

At the end, if you want to AuthN&Authz you need a user or a password, in this case certs. I would like to say that it's going to be easier in future, but if we're going to switch to FDO it's going to be even more complicated with 1-2 more servers. 

Having a dev profile to turn off it's hard, manly because the Authz is based on certificate data, so I'm not sure if we can have a developer workflow without compromise API security.

jakub-dzon (commented):
Please remove `HTTP_PORT` reference from controller_manager_config.properties

config/manager/manager.yaml
```diff
@@ -75,10 +73,6 @@ metadata:
     control-plane: controller-manager
 spec:
   ports:
-    - name: yggd
```
jakub-dzon, line 78: OCP route refers that port: https://github.com/project-flotta/flotta-operator/blob/f3906500370d89025ace44306439127cfa69bd15/config/ocp/network/route.yaml#L11

jakub-dzon (approved):

gciavarrini:
Does this PR affect [running.md](https://github.com/project-flotta/flotta-operator/blob/main/docs/user-guide/running.md) also?

PR #184, opened 2022-04-11
Add setup dev environment scripts

arielireni:
Created bash scripts for setup dev environments on OCP cluster and minikube.
Updated the `docs/user-guide/running.md` file in order to add documentation for using these scripts.

### DIFF
--- a/docs/user-guide/running.md
+++ b/docs/user-guide/running.md
@@ -16,6 +16,50 @@ When your environment is running, you can [deploy your workload](deploying-workl
 To force device re-registration remove EdgeDevice CR from the cluster and `/var/local/yggdrasil/device` directory from the device
 
 ---
+## Automatically
+
+### Prerequisites
+- Clone all Flotta components: [flotta-operator](https://github.com/project-flotta/flotta-operator), [flotta-device-worker](https://github.com/project-flotta/flotta-device-worker) and [yggdrasil](https://github.com/RedHatInsights/yggdrasil).
+- On minikube: add to `/etc/hosts`: `<minikube ip> <host name>`.
+
+### Start Flotta operator
+Script arguments: 
+- `IMG` (image registry and tag, for example: `IMG=quay.io/project-flotta/flotta-operator:latest`) 
+- `TARGET` (k8s or ocp) 
+- `OPERATOR_PATH` (optional)
+
+From the operator directory run:
+```
+bash ./hack/start-operator.sh -i <image> -t <target>
+```
+Note: if running outside the operator directory, add also `-o <operator-path>`.
+
+### Start Flotta agent
+Script arguments: 
+- `TARGET` (k8s or ocp)
+- `YGGDRASIL_PATH`
+- `DEVICE_WORKER_PATH` (optional)
+
+From the device-worker directory run:
+```
+bash <operator-path>/hack/start-agent.sh -t <target> -y <yggdrasil path> 
+```
+Note: if running outside the device-worker directory, pass also `-d <device-worker-path>`.
+
+### Start Flotta (both operator and agent)
+Script arguments:
+- `IMG` (image registry and tag, for example: `IMG=quay.io/project-flotta/flotta-operator:latest`)
+- `TARGET` (k8s or ocp)
+- `OPERATOR_PATH` 
+- `DEVICE_WORKER_PATH`
+- `YGGDRASIL_PATH`
+
+From the operator directory run:
+```
+bash ./hack/start-flotta.sh -i <image> -t <target> -o <operator-path> -d <device-worker-path> -y <yggdrasil path>
+```
+
+---
 
 ## Manually
 
--- a/hack/start-agent.sh
+++ b/hack/start-agent.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+usage()
+{
+cat << EOF
+Usage: $0 OPTIONS
+This script runs test for flotta-agent to verify it can start running properly.
+OPTIONS:
+   -d      Path to flotta-device-worker directory (optional)
+   -h      Show this message
+   -t      Target (ocp/k8s)
+   -y      Path to yggdrasil directory
+EOF
+}
+
+while getopts "t:y:d:" option; do
+    case "${option}"
+    in
+        t) TARGET=${OPTARG};;
+        y) YGGDRASIL_PATH=${OPTARG};;
+        d) DEVICE_WORKER_PATH=${OPTARG};;
+        h)
+            usage
+            exit 0
+            ;;
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+done
+
+if [[ -z $TARGET ]]; then
+    echo "Error: TARGET is required"
+    exit 1
+fi
+
+if [[ -z $YGGDRASIL_PATH ]]; then
+    echo "Error: YGGDRASIL_PATH is required"
+    exit 1
+fi
+
+if [[ -n $DEVICE_WORKER_PATH ]]; then
+  cd $DEVICE_WORKER_PATH
+fi
+
+sudo systemctl enable --now nftables.service
+sudo systemctl enable --now podman.service
+sudo systemctl enable --now podman.socket
+
+make uninstall
+make install
+
+cd $YGGDRASIL_PATH
+export PREFIX=/var/local
+make bin
+
+pid_file="${PREFIX}/var/run/yggdrasil/workers/device-worker.pid"
+if [ -f "$pid_file" ] ; then
+    sudo rm "$pid_file"
+fi
+
+if [[ "$TARGET" == "k8s" ]]; then
+  sudo ./yggd \
+    --protocol http \
+    --path-prefix api/flotta-management/v1 \
+    --client-id $(cat /etc/machine-id) \
+    --cert-file /tmp/cert.pem \
+    --key-file /tmp/key.pem \
+    --ca-root /tmp/ca.pem \
+    --log-level trace \
+    --server project-flotta.io:443
+
+elif [[ "$TARGET" == "ocp" ]]; then
+  HTTP_SERVER=$(oc get routes flotta-operator-controller-manager -n flotta --no-headers -o=custom-columns=HOST:.spec.host)
+  sudo ./yggd \
+    --log-level info \
+    --protocol http \
+    --path-prefix api/flotta-management/v1 \
+    --client-id $(cat /etc/machine-id) \
+    --cert-file /tmp/cert.pem \
+    --key-file /tmp/key.pem \
+    --server $HTTP_SERVER
+fi
+
+echo start-agent done!
--- a/hack/start-flotta.sh
+++ b/hack/start-flotta.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+usage()
+{
+cat << EOF
+Usage: $0 OPTIONS
+This script runs test for project-flotta to verify all flotta components can start running properly.
+OPTIONS:
+   -d      Path to flotta-device-worker directory
+   -h      Show this message
+   -i      Image registry
+   -o      Path to flotta-operator directory
+   -t      Target (ocp\k8s)
+   -y      Path to yggdrasil directory
+EOF
+}
+
+while getopts "i:t:o:d:y:" option; do
+    case "${option}"
+    in
+        i) IMG=${OPTARG};;
+        t) TARGET=${OPTARG};;
+        o) OPERATOR_PATH=${OPTARG};;
+        d) DEVICE_WORKER_PATH=${OPTARG};;
+        y) YGGDRASIL_PATH=${OPTARG};;
+        h)
+            usage
+            exit 0
+            ;;
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+done
+
+if [[ -z $OPERATOR_PATH ]]; then
+    echo "Error: OPERATOR_PATH is required"
+    exit 1
+fi
+
+if [[ -z $DEVICE_WORKER_PATH ]]; then
+    echo "Error: DEVICE_WORKER_PATH is required"
+    exit 1
+fi
+
+if [[ -z $YGGDRASIL_PATH ]]; then
+    echo "Error: YGGDRASIL_PATH is required"
+    exit 1
+fi
+
+if [[ -z $IMG ]]; then
+    echo "Error: IMG is required"
+    exit 1
+fi
+
+if [[ -z $TARGET ]]; then
+    echo "Error: TARGET is required"
+    exit 1
+fi
+
+sh hack/start-operator.sh -i $IMG -t $TARGET -o $OPERATOR_PATH
+if [ $? -eq 1 ]; then
+    echo "Error: start operator"
+else
+    sh hack/start-agent.sh -t $TARGET -y $YGGDRASIL_PATH -d $DEVICE_WORKER_PATH
+    if [ $? -eq 1 ]; then
+        echo "Error: start agent"
+    else
+        echo "Done!"
+    fi
+fi
--- a/hack/start-operator.sh
+++ b/hack/start-operator.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+usage()
+{
+cat << EOF
+Usage: $0 OPTIONS
+This script runs sanity check for project-flotta for testing flotta-operator is up and running properly.
+OPTIONS:
+   -h      Show this message
+   -i      Image registry
+   -o      Path to flotta-operator directory (optional)
+   -t      Target (ocp/k8s)
+EOF
+}
+
+while getopts "i:t:o:" option; do
+    case "${option}"
+    in
+        i) IMG=${OPTARG};;
+        t) TARGET=${OPTARG};;
+        o) OPERATOR_PATH=${OPTARG};;
+        h)
+            usage
+            exit 0
+            ;;
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+done
+
+if [[ -z $IMG ]]; then
+    echo "Error: IMG is required"
+    usage
+    exit 1
+fi
+
+if [[ -z $TARGET ]]; then
+    echo "Error: TARGET is required"
+    usage
+    exit 1
+fi
+
+if [[ -n $OPERATOR_PATH ]]; then
+  cd $OPERATOR_PATH
+fi
+
+sudo rm /tmp/*.pem
+make get-certs
+sudo chown root:root /tmp/*.pem
+IMG=$IMG make docker-build docker-push
+make install-router
+IMG=$IMG TARGET=$TARGET make deploy
+
+echo start-operator done!


eloycoto (commented):
I'm ok with the change, but I'm not sure if gnome-terminal is the best thing to add here

hack/start_flotta_k8s.sh
```diff
@@ -0,0 +1,39 @@
...
+        git checkout cdde836c519ac72302714b5f7ae90e45de7c79a7; \
+        export PREFIX=/var/local; \
+        make bin; \
+
+        sudo ./yggd \
+          --log-level info \
+          --protocol http \
+          --path-prefix api/flotta-management/v1 \
+          --client-id $(cat /etc/machine-id) \
+          --cert-file /etc/pki/consumer/cert.pem \
```
eloycoto, line 33: these certs are not correct, no? should be the ones from make get-certs

tupyy: +1. and don't forget the `sudo chown root.root /tmp/*.pem` pls

arielireni: Do you mean
```
--cert-file /tmp/cert.pem
--key-file /tmp/key.pem
```
?

hack/start_flotta_k8s.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+export OPERATOR_PATH="<path to flotta operator repo>"
```
jakub-dzon, line 3: If we deliver scripts, there should be no substitute values in them and should be usable without changes. 
The script should expect the user to specify env variables or script arguments and fail gracefully with a proper information if they are not provided.

hack/start_flotta_k8s.sh
```diff
@@ -0,0 +1,39 @@
...
+export DEVICE_WORKER_PATH="<path to flotta device worker repo>"
+export YGGDRASIL_PATH="<path to yggdrasil repo>"
+
+# Start flotta operator
+gnome-terminal --tab -- bash -c \
```
jakub-dzon, line 8: I don't think gnome-terminal is the best solution here. Consider `nohup` and smaller script files.
For example:
`nohup  > start-operator.sh >operator.log 2>&1 &`

jakub-dzon (changes requested):

hack/start_flotta_k8s.sh
gciavarrini: Is the `--ca-root /tmp/ca.pem` necessary?

hack/start_flotta_k8s.sh
```diff
@@ -0,0 +1,39 @@
...
+        "cd $DEVICE_WORKER_PATH; \
+        make uninstall; \
+        make install; \
+        exec bash"
+
+# Start yggdrasil
+gnome-terminal --tab -- bash -c \
+        "cd $YGGDRASIL_PATH; \
+        git checkout cdde836c519ac72302714b5f7ae90e45de7c79a7; \
```
gciavarrini, line 24: Doubt here: can we checkout the HEAD? In this way the script will always be up-to-date.

In addition [here](https://github.com/project-flotta/flotta-operator/blob/f3906500370d89025ace44306439127cfa69bd15/docs/user-guide/running.md?plain=1#L70) there isn't a specific commit to checkout.

docs/user-guide/running.md
```diff
@@ -15,6 +15,33 @@ When your environment is running, you can [deploy your workload](deploying-workl
 
 To force device re-registration remove EdgeDevice CR from the cluster and `/var/local/yggdrasil/device` directory from the device
 
+---
+## Automatically
+
+### Prerequisites
+- Clone all Flotta components: [flotta-operator](https://github.com/project-flotta/flotta-operator), [flotta-device-worker](https://github.com/project-flotta/flotta-device-worker) and [yggdrasil](https://github.com/RedHatInsights/yggdrasil).
+- Inside `hack/start_flotta_*.sh` define the image registry and paths to all the above components.
```
gciavarrini, line 23: Is it possible to use input parameters instead of editing the script?

hack/start_flotta_k8s.sh
arielireni: So what would be a better approach? define env variables before running the scripts (executing lines 3-5 manually before) or passing them as input parameters when running the script (and then running the script by `bash hack/start_flotta_k8s.sh path1 path2 path3 img`)?
(@gciavarrini related to your comment too)

hack/start_flotta_k8s.sh
```diff
@@ -0,0 +1,39 @@
...
+export YGGDRASIL_PATH="<path to yggdrasil repo>"
+
+# Start flotta operator
+gnome-terminal --tab -- bash -c \
+        "cd $OPERATOR_PATH; \
+        IMG=<image registry and tag> make docker-build docker-push; \
+        IMG=<image registry and tag> TARGET=k8s make deploy; \
+        exec bash"
+
+# Start flotta device worker
```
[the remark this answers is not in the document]
arielireni: Do you mean splitting to 4 scripts? (`start-operator.sh` , `start-device-worker.sh` , `start-yggdrasil.sh` and `start-all-flotta.sh`)

hack/start_flotta_k8s.sh
arielireni: If I remember correctly, after we started to use yggdrasil [here](https://github.com/RedHatInsights/yggdrasil), @eloycoto changed [here](https://github.com/project-flotta/flotta-device-worker/blob/main/README.md?plain=1#L54) to use this commit due to issues that came up. @eloycoto Can we checkout the HEAD or still need to use this commit?

In addition, @gciavarrini I think [this line](https://github.com/project-flotta/flotta-operator/blob/f3906500370d89025ace44306439127cfa69bd15/docs/user-guide/running.md?plain=1#L70) isn't updated this we use `RedHatInsights/yggdrasil` ([here](https://github.com/project-flotta/flotta-device-worker/blob/main/README.md?plain=1#L52))

hack/start_flotta_k8s.sh
gciavarrini: Honestly, I don't have a strong opinion on this.
Maybe, I prefer the script arguments with defaults to remote repo directory names (e.g. `../flotta-operator`, `../flotta-device-worker`, ... ) + check if they exist. I think arguments can be faster and easier to edit than env var.
But, once again, I don't know if this is the best approach. Let's wait for others' opinions.

hack/start_flotta_k8s.sh
arielireni: The main idea for using `gnome-terminal` here was to open automatically 3 tabs for the operator, device worker and yggdrasil. If I understand correctly `nohup` is not used for that purpose, no?

If we split this script anyway, another option is to pass the `start-all.sh` script and add to the documentation to run each script (`start-operator.sh` and `start-agent`) from a separate terminal (and then there's no need for `nohup` or `gnome-terminal`). WDYT?

(and BTW, why gnome-terminal isn't the best solution here?)

hack/start_flotta_k8s.sh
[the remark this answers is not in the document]
tupyy: What do you think about creating function inside the script and add a select case at root? In this way the user can use the same script for multiple commands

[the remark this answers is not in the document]
arielireni: What is the benefit of using this approach? since there isn't overlap between those scripts I'm not sure which way should be preferred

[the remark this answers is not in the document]
arielireni: Got you, thanks :)

arielireni:
Update:
- Now at the beginning of the script, the user will choose which Flotta component to start (operator, agent, or both).
- Then, the user will be requested to enter the required variables (one by one, depending on the chosen component).
- I removed the use of `gnome-terminal`, so now the whole script will run on the same terminal (and not three separate tabs as it was originally)

hack/start_flotta.sh
```diff
@@ -0,0 +1,72 @@
...
+
+  fi
+}
+
+start_all() {
+  start_operator
+  start_agent
+}
+
+read -p "Choose Flotta component to start: (1 - operator, 2 - agent, 3 - all) " OP
```
jakub-dzon, line 58: tl;dr; Please make it non-interactive. 

During dev I often tear down the env and set it up again; answering same questions every time will be very annoying and counter-productive. 

The flow I would like to follow:
 1. set-up env by providing all the input values in one command;
 2. tear down the dev
 3. set-up env by navigating command history and using command I used before (maybe changing one parameter)
 4. goto 2

jakub-dzon (changes requested):

hack/start_flotta.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+start_operator() {
+  read -p "Enter operator path: " OPERATOR_PATH
```
gciavarrini, line 4: This force the user to manually insert every time all the required paths. IMO it's not the best.

hack/start_flotta.sh
arielireni: So with this approach, I think it will be better to split the script since the arguments are different for each purpose

hack/start_flotta.sh
arielireni: So I split the script into three scripts :) each will receive different arguments

hack/start-flotta.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# arguments: $1 = operator path ; $2 = device-worker path ; $3 = yggdrasil path ; $4 = image registry ; $5 = target
```
gciavarrini, line 3: Overall, I'm ok with this PR (great job! :) ). 
My last doubt is about positional argument. It's easy to wrongly insert values but on the other hand the script will be more complicated and probably It will be an overkill.

tupyy: I would use arguments like in this script:
https://github.com/project-flotta/flotta-scale-tests/blob/main/scripts/run_test_plan.sh

arielireni: Agreed. That's why I added to `docs/user-guide/running.md` examples for each use case so the user will easily copy & paste the command and then change the values in place. WDYT?

arielireni: Done, @gciavarrini @tupyy  WDYT?

docs/user-guide/running.md
```diff
@@ -15,6 +15,55 @@ When your environment is running, you can [deploy your workload](deploying-workl
...
+---
+## Automatically
+
+### Prerequisites
+- Clone all Flotta components: [flotta-operator](https://github.com/project-flotta/flotta-operator), [flotta-device-worker](https://github.com/project-flotta/flotta-device-worker) and [yggdrasil](https://github.com/RedHatInsights/yggdrasil).
+- On minikube: add to `/etc/hosts`: `<minikube ip> <host name>`.
+
+### Start Flotta operator
+Script arguments: 
+- Image registry (for example: `IMG=quay.io/jdzon/flotta-operator:latest`)
```
masayag, line 27: pls use let's use quay.io/project-flotta/flotta-operator instead

docs/user-guide/running.md
```diff
@@ -15,6 +15,55 @@ When your environment is running, you can [deploy your workload](deploying-workl
...
+bash <operator-path>/hack/start-agent.sh TARGET=<target> YGGDRASIL_PATH=<yggdrasil path> 
+```
+Note: if running outside the device-worker directory, pass also `DEVICE_WORKER_PATH=<device-worker-path>`.
+
+### Start Flotta (both operator and agent)
+Script arguments:
+- Path to operator directory 
+- Path to device-worker directory
+- Path to yggdrasil directory
+- Image registry (for example: `IMG=quay.io/jdzon/flotta-operator:latest`)
```
masayag, line 54: pls use let's use quay.io/project-flotta/flotta-operator instead

docs/user-guide/running.md
arielireni: Done

docs/user-guide/running.md
arielireni: Done

hack/start-operator.sh
```diff
@@ -0,0 +1,29 @@
...
+if [[ -z $TARGET ]]; then
+    echo "Error: TARGET is required"
+    exit 1
+fi
+
+if [[ $# == 3 ]]; then
+  cd $OPERATOR_PATH
+fi
+
+IMG=$IMG make docker-build docker-push
```
jakub-dzon, line 26: If the values are exported at line 9, is IMG=$IMG and TARGET=$TARGET needed?

docs/user-guide/running.md
```diff
@@ -15,6 +15,55 @@ When your environment is running, you can [deploy your workload](deploying-workl
...
+
+From the operator directory run:
+```
+bash ./hack/start-operator.sh IMG=<image> TARGET=<target>
+```
+Note: if running outside the operator directory, pass also `OPERATOR_PATH=<operator-path>`.
+
+### Start Flotta agent
+Script arguments: 
+- Target (k8s or ocp)
```
jakub-dzon, line 39: I think that actually expected argument names need to be mentioned here.

docs/user-guide/running.md
```diff
@@ -15,6 +15,55 @@ When your environment is running, you can [deploy your workload](deploying-workl
...
+
+From the device-worker directory run:
+```
+bash <operator-path>/hack/start-agent.sh TARGET=<target> YGGDRASIL_PATH=<yggdrasil path> 
+```
+Note: if running outside the device-worker directory, pass also `DEVICE_WORKER_PATH=<device-worker-path>`.
+
+### Start Flotta (both operator and agent)
+Script arguments:
+- Path to operator directory
```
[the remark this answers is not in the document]
arielireni: Done

docs/user-guide/running.md
arielireni: Done

hack/start-operator.sh
arielireni: No, fixed

jakub-dzon (approved):

gciavarrini (approved):
lgtm

Great job @arielireni!

docs/user-guide/running.md
```diff
@@ -15,6 +15,55 @@ When your environment is running, you can [deploy your workload](deploying-workl
 
 To force device re-registration remove EdgeDevice CR from the cluster and `/var/local/yggdrasil/device` directory from the device
 
+---
+## Automatically
+
+### Prerequisites
```
pkliczewski, line 21: Do you think it makes sense to make sure that `nftables` and `podman` is installed to run the agent?

arielireni: As part of the documentation or the script?

hack/start-agent.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# arguments: TARGET, YGGDRASIL_PATH, (optional) DEVICE_WORKER_PATH
```
masayag, line 3: optional: Named arguments or using flags will be more descriptive and the parsing is more precise than positional approach, e.g. https://github.com/project-flotta/flotta-scale-tests/blob/main/scripts/run_test_plan.sh#L3

hack/start-agent.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+if [[ $# = 3 ]]; then
+  cd $DEVICE_WORKER_PATH
+fi
+make uninstall
+make install
+
+cd $YGGDRASIL_PATH
+git checkout cdde836c519ac72302714b5f7ae90e45de7c79a7
```
masayag, line 29: pls add a comment for this (we depend on a specific commit for device-worker to work)

hack/start-agent.sh
```diff
@@ -0,0 +1,61 @@
...
+export PREFIX=/var/local
+make bin
+sudo chown root.root /tmp/*.pem
```
masayag, line 32: should be `sudo chown root:root ...`

hack/start-agent.sh
```diff
@@ -0,0 +1,61 @@
...
+
+pid_file = "/var/local/var/run/yggdrasil/workers/device-worker.pid"
```
masayag, line 34: pls use PREFIX as prefix for this:
`pid_file="${PREFIX}/var/run/yggdrasil/workers/device-worker.pid"`
also, no spaces before and after the assignment operator

hack/start-agent.sh
```diff
@@ -0,0 +1,61 @@
...
+elif [[ "$TARGET" == "ocp" ]]; then
+  HTTP_SERVER=$(oc get routes flotta-operator-controller-manager -n flotta --no-headers -o=custom-columns=HOST:.spec.host)
+  sudo ./yggd \
+    --log-level info \
+    --protocol http \
+    --path-prefix api/flotta-management/v1 \
+    --client-id $(cat /etc/machine-id) \
+    --cert-file /tmp/cert.pem \
+    --key-file /tmp/key.pem \
+    --server $HOST_SERVER
```
masayag, line 58: s/HOST_SERVER/HTTP_SERVER

masayag (changes requested):

docs/user-guide/running.md
pkliczewski: I thought about the script but we could be explicit in the documentation as well.

hack/start-agent.sh
pkliczewski: can we use variable? Later we could move to specific tag/version

hack/start-agent.sh
arielireni: done, but I guess we'll change this again after PR #200 will be finished

hack/start-agent.sh
arielireni: done

hack/start-agent.sh
arielireni: Removed this line since now we can rebase on top of yggdrasil's main branch

hack/start-agent.sh
arielireni: done

arielireni:
This PR depends on PR #200 which should fix current issues with the ingress (and may cause further changes, such as running `make install-router` before deploying the operator on minikube)

hack/start-flotta.sh
```diff
@@ -0,0 +1,64 @@
...
+    exit 1
+fi
+
+if [[ -z $TARGET ]]; then
+    echo "Error: TARGET is required"
+    exit 1
+fi
+
+sh hack/start-operator.sh $IMG $TARGET $OPERATOR_PATH
+sh hack/start-agent.sh $TARGET $YGGDRASIL_PATH $DEVICE_WORKER_PATH
```
gciavarrini, line 62: I think the options are missing here.
If you execute:
```
./hack/start-flotta.sh -i quay.io/myrepos/flotta-operator:latest -t k8s -o . -d ../flotta-device-worker -y ../yggdrasil
```
I get 

```
Usage: hack/start-operator.sh OPTIONS
This script runs sanity check for project-flotta for testing flotta-operator is up and running properly.
OPTIONS:
   -i      Image registry
   -t      Target (ocp/k8s)
   -o      Path to flotta-operator directory (optional)
Error: TARGET is required
start-flotta done!
```
So adding the options should fix the problem:
```
sh hack/start-operator.sh -i $IMG -t $TARGET -o $OPERATOR_PATH
sh hack/start-agent.sh -t $TARGET -y $YGGDRASIL_PATH -d $DEVICE_WORKER_PATH
```

hack/start-flotta.sh
```diff
@@ -0,0 +1,64 @@
...
+
+echo start-flotta done!
```
gciavarrini, line 64: In addition can we avoid to print this if there were some errors in `  hack/start-operator.sh` and `hack/start-agent.sh ` (as in my previous comment for example)

docs/user-guide/running.md
```diff
@@ -15,6 +15,50 @@ When your environment is running, you can [deploy your workload](deploying-workl
...
+- Clone all Flotta components: [flotta-operator](https://github.com/project-flotta/flotta-operator), [flotta-device-worker](https://github.com/project-flotta/flotta-device-worker) and [yggdrasil](https://github.com/RedHatInsights/yggdrasil).
+- On minikube: add to `/etc/hosts`: `<minikube ip> <host name>`.
```
gciavarrini, line 23: WDYT to remember to add execution privilege for hack/start-*.sh scripts?

gciavarrini (changes requested):

hack/start-operator.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+usage()
+{
+cat << EOF
+Usage: $0 OPTIONS
+This script runs sanity check for project-flotta for testing flotta-operator is up and running properly.
+OPTIONS:
```
masayag, line 8: pls add also flag for help, e.g.
`-h      Show this message`

hack/start-operator.sh
```diff
@@ -0,0 +1,54 @@
...
+    exit 1
+fi
+
+if [[ -z $TARGET ]]; then
+    echo "Error: TARGET is required"
+    usage
+    exit 1
+fi
+
+if [[ $# == 6 ]]; then
```
masayag, line 44: you should check `OPERATOR_PATH` value instead

docs/user-guide/running.md
masayag: shouldn't the files have the correct permissions in github?

docs/user-guide/running.md
```diff
@@ -15,6 +15,50 @@ When your environment is running, you can [deploy your workload](deploying-workl
...
+
+### Start Flotta operator
+Script arguments: 
+- `IMG` (image registry and tag, for example: `IMG=quay.io/project-flotta/flotta-operator:latest`) 
+- `TARGET` (k8s or ocp) 
+- `OPERATOR_PATH` (optional)
+
+From the operator directory run:
+```
+bash ./hack/start-operator.sh -i <image>  -t <target>
```
masayag, line 33: nit: two spaces after `<image>`

docs/user-guide/running.md
```diff
@@ -15,6 +15,50 @@ When your environment is running, you can [deploy your workload](deploying-workl
 
 To force device re-registration remove EdgeDevice CR from the cluster and `/var/local/yggdrasil/device` directory from the device
 
+---
+## Automatically
```
eloycoto, line 19: TBH, I would move this file to here: 

https://github.com/project-flotta/project-flotta.github.io/blob/main/documentation/latest/dev/guide.md

hack/start-agent.sh
```diff
@@ -0,0 +1,86 @@
...
+   -h      Show this message
+   -t      Target (ocp/k8s)
+   -y      Path to yggdrasil directory
+EOF
+}
+
+while getopts "t:y:d:" option; do
+    case "${option}"
+    in
+        t) TARGET=${OPTARG};;
```
eloycoto, line 19: TBH: Agents shouldn't know if TARGET is ocp or k8s at all. At the end is only an API.

hack/start-agent.sh
```diff
@@ -0,0 +1,86 @@
...
+
+if [[ -n $DEVICE_WORKER_PATH ]]; then
+  cd $DEVICE_WORKER_PATH
+fi
+
+sudo systemctl enable --now nftables.service
+sudo systemctl enable --now podman.service
+sudo systemctl enable --now podman.socket
+
+make uninstall
```
eloycoto, line 51: In what folder this install/uninstall happens? I'm not sure If I got where.

hack/start-agent.sh
```diff
@@ -0,0 +1,86 @@
...
+    --protocol http \
+    --path-prefix api/flotta-management/v1 \
+    --client-id $(cat /etc/machine-id) \
+    --cert-file /tmp/cert.pem \
+    --key-file /tmp/key.pem \
+    --ca-root /tmp/ca.pem \
+    --log-level trace \
+    --server project-flotta.io:443
+
+elif [[ "$TARGET" == "ocp" ]]; then
```
eloycoto, line 74: From 74 to 84, you can delete it.

docs/user-guide/running.md
```diff
@@ -15,6 +15,50 @@ When your environment is running, you can [deploy your workload](deploying-workl
...
+- `YGGDRASIL_PATH`
+- `DEVICE_WORKER_PATH` (optional)
+
+From the device-worker directory run:
+```
+bash <operator-path>/hack/start-agent.sh -t <target> -y <yggdrasil path> 
+```
+Note: if running outside the device-worker directory, pass also `-d <device-worker-path>`.
+
+### Start Flotta (both operator and agent)
```
eloycoto, line 49: I'm not sure if this is for DEV or users, but I do not think that we should provide a solution like this for users or even devs.

hack/start-agent.sh
```diff
@@ -0,0 +1,86 @@
...
+cd $YGGDRASIL_PATH
+export PREFIX=/var/local
+make bin
+
+pid_file="${PREFIX}/var/run/yggdrasil/workers/device-worker.pid"
+if [ -f "$pid_file" ] ; then
+    sudo rm "$pid_file"
+fi
+
+if [[ "$TARGET" == "k8s" ]]; then
```
eloycoto, line 63: This if is no need at all.

hack/start-operator.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
eloycoto, line 1: Why would you run this script instead of make targets? I'm ok with this, but maybe we should use this script on kind workflow to have it all fully tested. ?

eloycoto (changes requested):

arielireni:
Due to the updated documentation [here](https://project-flotta.github.io/documentation/latest/gsg/minikube.html) and the [auto-gen scripts](https://github.com/project-flotta/flotta-operator/blob/main/Makefile#L288) for running the device locally, this task can be closed.

PR #186, opened 2022-04-13
change backoff with RetryWatcher

tupyy:
This PR solves [ECOPORJECT-454](https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=9673&view=detail&selectedIssue=ECOPROJECT-454).

<del>The `client-go` has been downgraded from latest version `0.23.5` because there are issues between `zap-logger@v1.15.0` and `k8s.io/apimachinery@v0.25.5`</del>


gabriel-farache:
Please rebase to be with go verion 1.17 that will solve your issue with go-client version

tupyy:
> Now that PR #163 is merged, please rebase to be with go verion 1.17 that will solve your issue with go-client version
 Ok for rebase but I don't think go.1.17 will solve the issue.

gabriel-farache:
> > Now that PR #163 is merged, please rebase to be with go verion 1.17 that will solve your issue with go-client version
> > Ok for rebase but I don't think go.1.17 will solve the issue.

With PR #163 the go-client dependency is also updated to the latest version along with its others dependencies, check the changes to see

gabriel-farache (changes requested):
Can you add tests? There is no tests for this feature and I think it would be good to have some
Especially since we only watch the LOG_LEVEL field in the configMap
Otherwise it seems good for me

go.mod
```diff
@@ -88,7 +88,7 @@ require (
 	github.com/prometheus/procfs v0.6.0 // indirect
 	github.com/sirupsen/logrus v1.8.1 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
-	github.com/stretchr/objx v0.2.0 // indirect
+	github.com/stretchr/objx v0.2.0 // indirect; indirct
```
gciavarrini, line 91: Is this change correct?

watchers/configmap_watcher.go
```diff
@@ -68,21 +67,21 @@ func checkConfigMapChanges(eventChannel <-chan watch.Event, dataField string, da
...
-		*watcher, innerErr = clientset.CoreV1().ConfigMaps(namespace).Watch(context.TODO(),
-			metav1.SingleObject(metav1.ObjectMeta{Name: configMapName, Namespace: namespace}))
-		if innerErr != nil {
-			logger.Info("cannot create watcher", "namespace", namespace, "configMap name", configMapName, "error", innerErr)
-		}
-		return innerErr
-	})
-	return err
+// watcher implements config.Watch interface which will create a ConfigMapWatcher.
+type watcher struct {
```
gciavarrini, line 71: WDYT to move this struct declaration below the `var` section? I think this can make the code style a bit more consistent.

watchers/configmap_watcher.go
```diff
@@ -68,21 +67,21 @@ func checkConfigMapChanges(eventChannel <-chan watch.Event, dataField string, da
...
 
-func retriable(err error) bool {
-	retry := err != nil
-	logger.V(1).Info("cannot create watcher", "retriable", retry, "error", err)
-	return retry
+func (w watcher) Watch(options metav1.ListOptions) (watch.Interface, error) {
+	watcher, err := w.clientset.CoreV1().ConfigMaps(w.namespace).Watch(context.Background(),
+		metav1.ListOptions{
+			ResourceVersion:     options.ResourceVersion,
+			AllowWatchBookmarks: options.AllowWatchBookmarks,
```
gciavarrini, line 81: I don't know how exactly bookmarks work (I red [this](https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks) ).
Last sentence says: 
> but you shouldn't assume bookmarks are returned at any specific interval, nor can clients assume that the API server will send any BOOKMARK event even when requested

Should we take care of this? (Asking to better understand the code)

tupyy: a little bit of context here:
the whole process changed with the RetryWatcher. Now the RetryWatcher wants a structure with some sort of creator watcher method which is called by the RetryWatcher when it starts. If you look at the code of the RetryWatcher (https://github.com/kubernetes/client-go/blob/28ccde769fc5519dd84e5512ebf303ac86ef9d7c/tools/watch/retrywatcher.go#L100) is like:
```go
func (rw *RetryWatcher) doReceive() (bool, time.Duration) {
	watcher, err := rw.watcherClient.Watch(metav1.ListOptions{
		ResourceVersion:     rw.lastResourceVersion,
		AllowWatchBookmarks: true,
	})
```
So, what I've done is take these arguments and inject them along with my configMap name into the `ListOptions` object passed to `configMapWatcher`. In this context, the `AllowWatchBookmarks` is always true.

watchers/configmap_watcher.go
tupyy: ok but this is private. I think the code is more clear with this structure here.
I don't have a strong opinion about it. 
I let you decide :)

go.mod
tupyy: no :)

watchers/configmap_watcher.go
gciavarrini: No strong opinion. Please do whatever you prefer :)

watchers/configmap_watcher.go
gciavarrini: Thanks!

tupyy:
> Can you add tests? There is no tests for this feature and I think it would be good to have some Especially since we only watch a specific field in the configMap, it would be good to be sure that we are actually only watching it and only it Otherwise it seems good for me

The problem with the test here is that the only thing this code does is to crash the pod. I don't know how you can test that :)

gabriel-farache:
> > Can you add tests? There is no tests for this feature and I think it would be good to have some Especially since we only watch a specific field in the configMap, it would be good to be sure that we are actually only watching it and only it Otherwise it seems good for me
> 
> The problem with the test here is that the only thing this code does is to crash the pod. I don't know how you can test that :)

Do not test that the pod is restarting, we can assume that the exit command is working fine and that K8S will also restart the pod once it dies (it would be integration test if you were testing the pods restart) 

But you can test that there is a log message "restarting pod to update the logging level" t(or something like that) which is actually logged by the watcher function, with this, you can test that if the watched parameter is changed, something happen else nothing happen

go.mod
jakub-dzon: Please fix it;)

watchers/configmap_watcher.go
jakub-dzon: I think it should be moved to the top; structs with structs, methods later.

jakub-dzon (commented):
LGTM; please address the comments.

tupyy:
marking this as WIP cause there are errors in the cluster due to the retry:
```
E0519 08:48:47.167059       1 retrywatcher.go:130] "Watch failed" err="the server rejected our request for an unknown reason (get configmaps)"
```

watchers/configmap_watcher_test.go
```diff
@@ -0,0 +1,153 @@
...
+			Object: &to,
+		}
+
+		// wait a bit
+		<-time.After(2 * time.Second)
+
+		Expect(exitCallsCount).To(Equal(0))
+	})
+
+	AfterEach(func() {
```
eloycoto, line 150: we use to have BeforeEach and AfterEach just together it's just easy to read/follow

watchers/configmap_watcher_test.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		// wait a bit
+		<-time.After(2 * time.Second)
+
+		Expect(exitCallsCount).To(Equal(1))
+	})
+
+	It("do nothing after Delete event", func() {
+		go func() {
```
eloycoto, line 125: This go func can be part of the BeforeEAch no?

watchers/configmap_watcher_test.go
```diff
@@ -0,0 +1,153 @@
...
+		logger = zap.New()
+		exitCallsCount = 0
+
+		configMapMock.EXPECT().Watch(gomock.Any(), gomock.Any()).Return(watcherMock, nil).AnyTimes()
+		watcherMock.EXPECT().ResultChan().Return(resultCh).AnyTimes()
+		watcherMock.EXPECT().Stop().AnyTimes()
+	})
+
+	It("do nothing when initialFieldValue is not modified following an Add event", func() {
+		go func() {
```
eloycoto, line 47: Even though I do not really liked it, I would keep the way to do test like // given // when //then. comments to keep is consistent across project.

watchers/configmap_watcher.go
```diff
@@ -3,44 +3,47 @@ package watchers
...
-		checkConfigMapChanges(watcher.ResultChan(), dataField, dataValue)
+	wc := watcher{
+		configMapGetter: cmi,
+		configMapName:   configMapName,
 	}
+
+	w, err := clientwatch.NewRetryWatcher(lastResourceVersion, wc)
+	if err != nil {
+		logger.Error(err, "cannot create watcher", "configMap name", configMapName)
+		os.Exit(1)
```
eloycoto, line 40: shouldn't we use here the exitfunc()?

eloycoto: and maybe this should be also tested.

eloycoto (commented):
@jakub-dzon please review this one, I'm not sure if the operator split any issues with these watchers.

watchers/configmap_watcher.go
tupyy: do you test every `os.Exit(1)` in the project? especially in `main`?

tupyy: what do you mean not using at all? I'm using it.
Ok. whatever..

pkliczewski:
@jakub-dzon @masayag please take a look

watchers/configmap_watcher.go
tupyy: Actually, I realized that using exitFunc at this point is not safe. If, for some reason, the `exitFunc` does not contain a syscall to exit the flow will continue but with a `nil` watcher which will cause panic. 
I'm pushing the changes but I'm not really ok with it.

eloycoto: I'm not sure if I follow you here, the `exitfunc` should work also in all other calls, so it's no longer safe to use?

tupyy: What I wanted to say is that we leave the door open for a panic. If for some reason, the `exitFunc` passed to `WatchForChanges` does not contain an exit call and the `retyWatcher` fails to create the watcher then you got a panic.
But maybe it is just a theoretical case, I admit

jakub-dzon:
@tupyy - please make the build green.

eloycoto:
Maybe will be good to squash the commits for this PR.

pkliczewski:
@gabriel-farache please take a look whether this PR meets your expectations

gabriel-farache (approved):

gabriel-farache:
\lgtm

PR #187, opened 2022-04-13
Fix: kind workflow ignores condition error

arielireni:
The condition:
```
timeout 1m make deploy IMG=flotta-operator || retVal=$?
if [ $retVal -ne 124 ];
 then
    echo "Make run failed"; exit 1
fi
```
produced the following error in the logs when the build passes successfully:
`-ne: unary operator expected`

This error was caused since we should wrap the condition with one more `[]` and also check if `$retVal` is set before we check if his value is not 124. 
Therefore, the solution is to replace the condition with: `if [[ -n $retVal && $retVal -ne 124 ]]; then`
 
### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -34,7 +34,7 @@ jobs:
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
           timeout 2m make deploy IMG=flotta-operator TARGET=kind || retVal=$?
-          if [ $retVal -ne 124 ]; then
+          if [[ -n "$retVal" && $retVal -ne 124 ]]; then
             echo "Make run failed"; exit 1
           fi
           kubectl wait --timeout=120s --for=condition=Ready pods --all -n flotta


arielireni:
@machacekondra Please review

.github/workflows/kind.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
           timeout 2m make deploy IMG=flotta-operator TARGET=kind || retVal=$?
-          if [ $retVal -ne 124 ]; then
+          if [ "$retVal" -ne "124" ]; then
```
gciavarrini, line 37: Did you try ` if [[ $retVal -ne 124 ]]; then` ?
If i remember correctly `-ne` only works with ``[[``

(Not sure :) )

arielireni: I tried that now, but with this syntax the condition is true (`$retVal != 124` is true), so I received `Make run failed` and kind exits (even though it shouldn't)

gciavarrini: IMO, `retVal` is unset.

if `timeout 2m make deploy IMG=flotta-operator TARGET=kind` doesn't fail, then the `retVal=$?` isn't executed (there is the `||` operator! ).
So ` $retVal -ne 124` is true and that's correct.

You can try
```
> false || echo $?
1
> true || echo $?
```

and

```
#!/bin/bash

retVal=123
if [[ $retVal -ne 124 ]]
    then
        echo "Then branch"
    else
        echo "Else branch"
fi

unset retVal
if [[ $retVal -ne 124 ]]
    then
        echo "Then branch"
    else
        echo "Else branch"
if
```

It returns:
```
> ./test.sh
Then branch
Then branch
```
In addition, `-ne` shouldn't be used with strings (see [bash conditional expression](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html), [here](https://www.delftstack.com/howto/linux/bash-ne-not-equal/)  or [here](https://linuxtect.com/linux-bash-not-equal-ne-operators-tutorial/) for example)

arielireni: Nice! so I'll update the condition to check whether `retVal` is set and only if it is, it will check its value:
`if [[ -n $retVal && $retVal -ne 124 ]]; then`

gciavarrini: I'm sorry to bother you again on this but can you please check if `-n` works with non-string?
Since `-n` checks if a string is non-empty, I think we should use `-n "$retVal"` instead of `-n $retVal`.
(I'm still refering on [Bash conditional expression](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html))

.github/workflows/kind.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
           timeout 2m make deploy IMG=flotta-operator TARGET=kind || retVal=$?
-          if [ $retVal -ne 124 ]; then
+          if [[ -n $retVal && $retVal -ne 124 ]]; then
```
gciavarrini, line 37: I'm sorry to bother you again on this but can you please check if `-n` works with non-string?
Since `-n` checks if a string is non-empty, I think we should use `-n "$retVal"` instead of `-n $retVal`.
(I'm still refering on [Bash conditional expression](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html))

arielireni: `-n` is the opposite of `-z` which is used similarly [here](https://github.com/project-flotta/flotta-scale-tests/blob/main/scripts/run_test_plan.sh#L79), but I changed this to `-n "$retVal"` and it works too if you prefer :)

masayag (approved):

pkliczewski (approved):

PR #191, opened 2022-04-18
Approval process phase 2

eloycoto:
Three different commits: 

1) Auto-approval process: where by default, all requests to register will auto-approve by default. 
2) Update registration flow in the API, so no new edge devices will be created, certs are mandatory, etc..
3) Add a way to set the deviceSet on the EdgeDeviceSignedRequest.


eloycoto:
Device worker fails because needs this one: 
https://github.com/project-flotta/flotta-device-worker/pull/128

internal/yggdrasil/yggdrasil.go
```diff
@@ -299,73 +299,59 @@ func (h *Handler) PostDataMessageForDevice(ctx context.Context, params yggdrasil
...
-			return operations.NewPostDataMessageForDeviceOK().WithPayload(&res)
+		isInit := false
+		if dvc.ObjectMeta.Labels[v1alpha1.EdgeDeviceSignedRequestLabelName] == v1alpha1.EdgeDeviceSignedRequestLabelValue {
+			isInit = true
 		}
 
-		if !errors.IsNotFound(err) {
-			return operations.NewPostDataMessageForDeviceInternalServerError()
+		// the first time that tries to register should be able to use register certificate.
+		if !isInit && !IsOwnDevice(ctx, deviceID) {
```
pkliczewski, line 321: Is it possible for this condition to be used when certificate expires and the device is waiting to registration to be approved?

controllers/edgedevicesignedrequest_controller_test.go
```diff
@@ -99,7 +99,7 @@ var _ = Describe("EdgeDeviceSignedRequest controller", func() {
 		Expect(result.Requeue).To(BeFalse())
 	})
 
-	It("Device Signed request is not yet approved", func() {
+	FIt("Device Signed request is not yet approved", func() {
```
masayag, line 102: pls note there is focus on this test :-)

internal/yggdrasil/yggdrasil.go
eloycoto: yes, that logic happens on `mtls.VerifyRequest`

pkliczewski: do I understand correctly that if registration cert has expired and autoapprove process is not enabled the device won't be able to register?

api/v1alpha1/edgedevicesignedrequests_types.go
```diff
@@ -36,6 +36,9 @@ type EdgeDeviceSignedRequestSpec struct {
 	// TargetNamespace is the namespace where device will land
 	TargetNamespace string `json:"target_namespace"`
 
+	// TargetSet is set that edgedevice will use.
+	TargetSet string `json:"target_set,omitempty"`
```
jakub-dzon, line 40: I haven't paid enough attention before - the convention of json properties for CRDs is camelCase; please adjust `EdgeDeviceSignedRequest` (that includes `target_namespace`).

main.go
```diff
@@ -123,6 +123,9 @@ var Config struct {
 
 	// MaxConcurrentReconciles is the maximum number of concurrent Reconciles which can be run
 	MaxConcurrentReconciles uint `envconfig:"MAX_CONCURRENT_RECONCILES" default:"3"`
+
+	// AutoApprovalprocess enable auto approval on devices
+	AutoApproval bool `envconfig:"AUTO_APPROVAL_PROCESS" default:"true"`
```
jakub-dzon, line 128: I've got doubts here about `true` being the default. IMO the default should be what user is recommended to use and auto approval should be an exceptional case.

eloycoto: Mainly because of the use cases, when someone uses for the first time, it's easy to get the workload running without much effort. It's more to reduce the friction, something which I think that it's needed.

masayag (approved):

PR #192, opened 2022-04-18
Docs: Fix issues with running guide

eloycoto:

### DIFF
--- a/docs/user-guide/running.md
+++ b/docs/user-guide/running.md
@@ -63,11 +63,7 @@ sudo chown root:root /tmp/*.pem
 
 #### Prerequisites
 
-- Generate certificate and a key:
-
-  `openssl req -new -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -out cert.pem -keyout key.pem`
-- Put generated cert.pem and key.pem files in `/etc/pki/consumer` edge device directory 
-- Checkout https://github.com/jakub-dzon/yggdrasil repository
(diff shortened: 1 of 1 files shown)


docs/user-guide/running.md
```diff
@@ -63,11 +63,7 @@ sudo chown root:root /tmp/*.pem
 
 #### Prerequisites
 
-- Generate certificate and a key:
-
-  `openssl req -new -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -out cert.pem -keyout key.pem`
-- Put generated cert.pem and key.pem files in `/etc/pki/consumer` edge device directory 
-- Checkout https://github.com/jakub-dzon/yggdrasil repository
+- Checkout https://github.com/RedHatInsights/yggdrasil repository on main branch
```
masayag, line 66: so there is no longer need for the specific commit as mentioned in https://github.com/project-flotta/flotta-device-worker#start-yggdrasil ?

eloycoto: latest main will work, and we're speaking about the next stable branch

masayag (approved):

jakub-dzon (approved):

docs/user-guide/running.md
pkliczewski: I would not use latest. There could be regressions introduced which would cause us to waste some time. Let's try to pin ourselves to something we know works.

PR #200, opened 2022-04-27
Ingress Changes

eloycoto:
I was troubleshooting `nginx-ingress` and the TLS passthrough option on Minikube was a bit painful to set. So, for now, I switched the ingress to the Openshift router, where the TCP load-balancing works as expected and things are easier to debug. 

To install: 

```
make install-router
make deploy IMG=quay.io/project-flotta/flotta-operator TARGET=k8s
```

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -22,6 +22,9 @@ jobs:
         with:
           go-version: 1.17
 
+      - uses: satackey/action-docker-layer-caching@v0.0.11
+        continue-on-error: true
+
       - uses: engineerd/setup-kind@v0.5.0
         with:
           version: "v0.11.1"


config/network/route.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
   name: controller-manager
   namespace: system
 spec:
+  host: REPLACE_HOST
```
masayag, line 10: iinw, in OCP this value is decided by the ingress-operator (service-name-namespace-name-cluster-name)

eloycoto: but that value should be in alignment with the flotta/CA certs.

.github/workflows/kind.yaml
```diff
@@ -22,35 +22,32 @@ jobs:
...
           kubectl wait --for=condition=Ready pods --all -n kube-system
+          make install-router
 
           # Deploy flotta operator
           sed -i 's/LOG_LEVEL=info/LOG_LEVEL=debug/g' config/manager/controller_manager_config.properties
           make build
           IMG=flotta-operator:latest make docker-build
           kind load docker-image flotta-operator:latest
-          timeout 2m make deploy IMG=flotta-operator TARGET=kind || retVal=$?
+          timeout 2m make deploy IMG=flotta-operator TARGET=k8s || retVal=$?
```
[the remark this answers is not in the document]
eloycoto: this PR is marked as WIP, not yet finished ;-)

config/network/route.yaml
masayag: so in that case, we should pick the name when creating the cert based on expected host name, no?

eloycoto: yeah, it's a parameter on the operator config, default is project-flotta.io and 127.0.0.1

gciavarrini:
No need to update the documentation?

eloycoto:
> No need to update the documentation?

Just a follow-up PR in docs as soon this is finished :-)

gciavarrini (approved):
lgtm

machacekondra (approved):

config/network/route.yaml
arielireni: where should we pass the host here?
I tried to start yggdrasil with:
`sudo ./yggd --protocol http --path-prefix api/flotta-management/v1 --client-id $(cat /etc/machine-id) --cert-file /tmp/cert.pem --key-file /tmp/key.pem --ca-root /tmp/ca.pem --server 127.0.0.1:8043`
but received connection refused

pkliczewski (approved):

PR #209, opened 2022-05-06
Controller: remove edgedevice signed request when device is deleted

eloycoto:
When the device is removed, the EdgeDevice Signed request is still
approved, but the controller will never create a new entry for device,
so it keeps in a loop of no option to re-register.

For now, if a device is deleted, the EDSR will also be deleted.

Fix ECOPROJECT-776

### DIFF
--- a/controllers/edgedevice_controller.go
+++ b/controllers/edgedevice_controller.go
@@ -26,6 +26,7 @@ import (
 	obv1 "github.com/kube-object-storage/lib-bucket-provisioner/pkg/apis/objectbucket.io/v1alpha1"
 	"github.com/project-flotta/flotta-operator/internal/metrics"
 	"github.com/project-flotta/flotta-operator/internal/repository/edgedevice"
+	"github.com/project-flotta/flotta-operator/internal/repository/edgedevicesignedrequest"
 	"github.com/project-flotta/flotta-operator/internal/storage"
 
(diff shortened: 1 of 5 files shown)


pkliczewski (commented):
/lgtm

machacekondra (approved):

jakub-dzon (approved):

tupyy (approved):

controllers/edgedevice_controller.go
```diff
@@ -78,6 +81,18 @@ func (r *EdgeDeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	if edgeDevice.DeletionTimestamp != nil {
+		edsr, err := r.EdgeDeviceSignedRequestRepository.Read(ctx, edgeDevice.Name, r.InitialDeviceNamespace)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{Requeue: true}, err
```
masayag, line 90: the `Requeue: true` will not be considered if an error occurred and can be skipped:

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/controller.go#L322

PR #211, opened 2022-05-06
Improve e2e tests

machacekondra:
- validate systemd service is running
- add flag to skip pulling of images
- add test for different name of pod and container


machacekondra:
This should fail, with this PR merged should succeed: https://github.com/project-flotta/flotta-device-worker/pull/144

masayag (approved):

PR #212, opened 2022-05-09
Agent installation adding port option

eloycoto:
When running agent script always uses port 8043, with this patch port
8043 is set by default, but you can run with a specific argument:

```
$ hack/install-agent-dnf.sh -i 192.168.2.10 -p 443
```

### DIFF
--- a/hack/install-agent-dnf.sh.template
+++ b/hack/install-agent-dnf.sh.template
@@ -7,19 +7,23 @@ Usage: $0 OPTIONS
 This script installs Flotta agent component on RPM-compatible machine.
 OPTIONS:
    -h      Show this message
-   -i      IP address under which your Flotta operator API endpoint on port 8043 is accessible to the device.
-           When your KUBECONFIG is pointing to cluster with Flotta operator deployed, you can run
+   -i      IP address under which your Flotta edge API endpoint is accessible to the device.
+           For testing, when your KUBECONFIG is pointing to cluster with Flotta operator deployed, you can run
 
            kubectl port-forward service/flotta-operator-controller-manager -n flotta 8043 --address 0.0.0.0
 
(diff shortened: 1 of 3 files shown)


hack/install-agent-dnf.sh.template
```diff
@@ -13,13 +13,17 @@ OPTIONS:
            kubectl port-forward service/flotta-operator-controller-manager -n flotta 8043 --address 0.0.0.0
```
gabriel-farache, line 13: change `8043` by `<FLOTTA_PORT>` or something like that to emphasis that the provided port should be used. Or add `ie: ` before the command to be clear that it's a simple example using default values.
Also, change [this](https://github.com/project-flotta/flotta-operator/blob/feb03938c5fd901b2db04904d01375735d4e3919/hack/install-agent-dnf.sh.template#L10) that describe `-i` option

hack/install-agent-rpm-ostree.sh.template
```diff
@@ -13,13 +13,15 @@ OPTIONS:
            kubectl port-forward service/flotta-operator-controller-manager -n flotta 8043 --address 0.0.0.0
```
gabriel-farache, line 13: change 8043 by <FLOTTA_PORT> or something like that to emphasis that the provided port should be used. Or add ie: before the command to be clear that it's a simple example using default values.
Also, change [this](https://github.com/project-flotta/flotta-operator/blob/feb03938c5fd901b2db04904d01375735d4e3919/hack/install-agent-rpm-ostree.sh.template#L10) that describe -i option

gabriel-farache (changes requested):

gabriel-farache (approved):

masayag (approved):

PR #217, opened 2022-05-10
Makefile: Add cert-manager target

eloycoto:


eloycoto:
This change is for the development guide doc.

PR #219, opened 2022-05-11
Doc: Remove docs and point to the web

eloycoto:
As agreed in the meetings. 

Just to avoid having documentation in different places, have an entry place where all documentation is located, that in this case is the web. 


pkliczewski (commented):
/lgtm

masayag (approved):

PR #225, opened 2022-05-15
Reused target for pre-build actions

masayag:
The pre-build actions are grouped into a single target `pre-build`
and reused when needed to produce the same output when building
the project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ lint: ## Check if the go code is properly written, rules are in .golangci.yml
 ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 test: ## Run tests.
 test: GINKGO_OPTIONS ?= --skip e2e
-test: manifests generate fmt vet imports test-fast
+test: manifests pre-build test-fast
 
 integration-test: ginkgo get-certs
 ifeq ($(SKIP_TEST_IMAGE_PULL), false)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -146,13 +146,16 @@ check-certs: # Check cert subject
...
 ##@ Build
-build: generate fmt vet imports ## Build manager binary.
+pre-build: ## Generate code, format it and organize imports before executing build
+pre-build: generate fmt imports vet
+
+build: pre-build ## Build manager binary.
 	go build -mod=vendor -o bin/manager main.go
 
-fast-build: generate fmt vet ## Fast build manager binary for local dev.
+fast-build: pre-build ## Fast build manager binary for local dev.
```
eloycoto, line 155: maybe is not fast-build at all :joy:

eloycoto (approved):

Makefile
```diff
@@ -146,13 +146,16 @@ check-certs: # Check cert subject
...
 ##@ Build
-build: generate fmt vet imports ## Build manager binary.
+pre-build: ## Generate code, format it and organize imports before executing build
+pre-build: generate fmt imports vet
+
+build: pre-build ## Build manager binary.
 	go build -mod=vendor -o bin/manager main.go
 
-fast-build: generate fmt vet ## Fast build manager binary for local dev.
+fast-build: generate ## Fast build manager binary for local dev.
```
masayag, line 155: @eloycoto changed this a bit to perform minimal steps for compilation and building the binary.
PTAL If you think it should be removed or changed.

jakub-dzon (approved):

Makefile
eloycoto: it's ok, it's ok :-)

PR #227, opened 2022-05-16
Added metrics for authN

jordigilh:
* flotta_operator_edge_devices_failed_authentication: counts the number of
times a device fails to authenticate, such as invalid cert, expired
cert, no cert.

* flotta_operator_edge_devices_invalid_owner: counts the number of times
  a device ID mismatches the ID stored in the context of the request.

### DIFF
--- a/internal/yggdrasil/yggdrasil.go
+++ b/internal/yggdrasil/yggdrasil.go
@@ -142,6 +142,7 @@ func (h *Handler) getNamespace(ctx context.Context) string {
 func (h *Handler) GetControlMessageForDevice(ctx context.Context, params yggdrasil.GetControlMessageForDeviceParams) middleware.Responder {
 	deviceID := params.DeviceID
 	if !IsOwnDevice(ctx, deviceID) {
+		h.metrics.IncEdgeDeviceInvalidOwnerCounter()
 		return operations.NewGetControlMessageForDeviceForbidden()
 	}
 	logger := log.FromContext(ctx, "DeviceID", deviceID)
@@ -172,6 +173,7 @@ func (h *Handler) GetControlMessageForDevice(ctx context.Context, params yggdras
 func (h *Handler) GetDataMessageForDevice(ctx context.Context, params yggdrasil.GetDataMessageForDeviceParams) middleware.Responder {
 	deviceID := params.DeviceID
 	if !IsOwnDevice(ctx, deviceID) {
+		h.metrics.IncEdgeDeviceInvalidOwnerCounter()
 		return operations.NewGetDataMessageForDeviceForbidden()
 	}
 	logger := log.FromContext(ctx, "DeviceID", deviceID)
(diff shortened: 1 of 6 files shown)


internal/yggdrasil/yggdrasil.go
```diff
@@ -142,6 +142,7 @@ func (h *Handler) getNamespace(ctx context.Context) string {
 func (h *Handler) GetControlMessageForDevice(ctx context.Context, params yggdrasil.GetControlMessageForDeviceParams) middleware.Responder {
 	deviceID := params.DeviceID
 	if !IsOwnDevice(ctx, deviceID) {
+		h.metrics.IncEdgeDeviceInvalidOwnerCounter()
 		return operations.NewGetControlMessageForDeviceForbidden()
 	}
```
masayag, line 147: this block repeats itself 4 times.
but this is one of that cases in which refactoring into a function will not save LOC, but will allow to encapsulate:
* Owner validation
* Increase metrics counter
* Report the same error for this violation

however, I'm not strongly pushing to it. just an idea. going to give my LGTM regardless.

internal/yggdrasil/yggdrasil.go
```diff
@@ -142,6 +142,7 @@ func (h *Handler) getNamespace(ctx context.Context) string {
 func (h *Handler) GetControlMessageForDevice(ctx context.Context, params yggdrasil.GetControlMessageForDeviceParams) middleware.Responder {
 	deviceID := params.DeviceID
 	if !IsOwnDevice(ctx, deviceID) {
+		h.metrics.IncEdgeDeviceInvalidOwnerCounter()
```
masayag, line 145: this block (lines 144-147) repeats itself 4 times.
but this is one of that cases in which refactoring into a function will not save LOC, but will allow to encapsulate:

Owner validation
Increase metrics counter
Report the same error for this violation
however, I'm not strongly pushing to it. just an idea. going to give my LGTM regardless.

masayag (approved):

masayag:
@jordigilh pls run `make all` locally again before pushing.
It seems that the output of `make imports` is different than the expected one.

pkliczewski (commented):
/lgtm, please fix CI job

internal/yggdrasil/yggdrasil.go
jordigilh: Does not repeat entirely, the return object is different:

First instance we return a control message forbidden:
https://github.com/project-flotta/flotta-operator/blob/8dc343aae72560ff35ee060d7620e44d1db6a933/internal/yggdrasil/yggdrasil.go#L144-L146

Here we return a data message forbidden:
https://github.com/project-flotta/flotta-operator/blob/8dc343aae72560ff35ee060d7620e44d1db6a933/internal/yggdrasil/yggdrasil.go#L175-L178

And for the last 2 ones we return a post data message forbidden:

https://github.com/project-flotta/flotta-operator/blob/8dc343aae72560ff35ee060d7620e44d1db6a933/internal/yggdrasil/yggdrasil.go#L218-L221

But I see your point. However, even if they were the same object, I'd still have concerns about readability by encapsulating the metric call in this case, as the other checks in this method wouldn't follow the same trend. Doing it here would create a disparity level in terms of readability/visibility of how the function is read since you don't have to drill down to understand what it does.

This is one case where readability would overrule code replication in my opinion.

jakub-dzon (approved):

pkliczewski (approved):

PR #235, opened 2022-05-24
Data-configuration: introduce ingress/egress data slices

jordigilh:
Changes in the swagger for data-configuration to allow specifying egress (existing) or ingress data synchronization on the worker.


internal/yggdrasil/configuration-assembler.go
```diff
@@ -230,12 +230,21 @@ func (a *configurationAssembler) toWorkloadList(ctx context.Context, logger *zap
...
+			if len(spec.Data.Egress) > 0 {
+				var e []*models.DataPath
+				for _, path := range spec.Data.Egress {
+					e = append(e, &models.DataPath{Source: path.Source, Target: path.Target})
+				}
+				data.Egress = e
+			}
+			if len(spec.Data.Ingress) > 0 {
+				var i []*models.DataPath
+				for _, path := range spec.Data.Egress {
```
gabriel-farache, line 243: Should it be` spec.Data.Igress` instead of` spec.Data.Egress`

internal/yggdrasil/configuration-assembler.go
```diff
@@ -230,12 +230,21 @@ func (a *configurationAssembler) toWorkloadList(ctx context.Context, logger *zap
 			continue
 		}
 		var data *models.DataConfiguration
-		if spec.Data != nil && len(spec.Data.Paths) > 0 {
-			var paths []*models.DataPath
-			for _, path := range spec.Data.Paths {
-				paths = append(paths, &models.DataPath{Source: path.Source, Target: path.Target})
+		if spec.Data != nil {
+			if len(spec.Data.Egress) > 0 {
```
gabriel-farache, line 234: Is the if necessary? The `range` on empty slice will not do anything and `e` is nil at this point the the `data.Egress = e` after will not have an impact. I'm not sure what is the go-ic (or golangic (like pythonic)) way 
Maybe you can create separate methods/functions; one for each egress and ingress creation

internal/yggdrasil/configuration-assembler.go
jordigilh: Thanks!

gabriel-farache (approved):

PR #240, opened 2022-05-27
Swagger backend REST API

jordigilh:
Swagger definition for the backend REST API defined in https://github.com/project-flotta/flotta-operator/pull/236


swagger-backend.yaml
```diff
@@ -0,0 +1,191 @@
...
+        type: boolean
+  device-registration-response:
+    type: object
+    properties:
+      target-namespace:
+        type: string
+      message:
+        type: string
+paths:
+  '/namespaces/{namespace}/devices/{device-id}/deregister':
```
jakub-dzon, line 42: I don't think that `namespaces` path segment is needed.

jakub-dzon, line 42: `.../deregistrable` or `.../finalize` (if you agree with [my comment](https://github.com/project-flotta/flotta-operator/pull/236/files/d712fed91ac4c26188efc552de3230ba6123160d#r884143325)

jakub-dzon: Or maybe general path pattern like `/devices/{namespace}/{device-id}/...`?

swagger-backend.yaml
```diff
@@ -0,0 +1,191 @@
...
+          type: string
+          required: true
+        - in: body
+          name: enrolment-info
+          required: true
+          schema:
+            $ref: 'swagger.yaml#/definitions/enrolment-info'
+      responses:
+        '200':
+          description: Success
```
jakub-dzon, line 118: We need to somehow encode in the response that the device is already present - either in the body or status code. Maybe HTTP 201 (Created) when new enrollment is recoded and HTTP 200 (OK) when the device already exists?

swagger-backend.yaml
```diff
@@ -0,0 +1,191 @@
...
+            $ref: '#/definitions/device-registration-response'
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/device-registration-response'
+    put:
```
jakub-dzon, line 160: :+1:

swagger-backend.yaml
```diff
@@ -0,0 +1,191 @@
...
+          description: Namespace where the device resides
+          type: string
+          required: true
+        - in: path
+          name: device-id
+          description: Device ID
+          type: string
+          required: true
+        - in: body
+          name: registration-information
```
jakub-dzon, line 143: That should be for `put` and what's in `put` should be here.

jakub-dzon: Scratch that...

jakub-dzon:
The heartbeat operation is missing.

swagger-backend.yaml
jordigilh: that would be a bit confusing, as the `{namespace}` could be interepreted as being the `{device-id}` and vice versa. I'd prefer to either remove the namespace and put it as part of the payload or keep it as it is.

jordigilh: I'll take `/finalize` :smile:

swagger-backend.yaml
jordigilh: Sounds good. Thanks for the suggestion.

swagger-backend.yaml
jakub-dzon: OK. That's fine.

jordigilh:
> The heartbeat operation is missing.

Thanks, I added the `UpdateHearbeat` endpoint.

swagger-backend.yaml
jordigilh: s/POST/PUT/ I realized that semantically doesn't make sense to do a POST if you haven't done a PUT first. So PUT it is.

jakub-dzon (commented):
The swagger design LGTM. 
Please get rid of all the errors.

Makefile
```diff
@@ -275,6 +275,7 @@ k8s-client-gen: client-gen lister-gen informer-gen ## Generate typed client for
 
 validate-swagger: ## Validate swagger
 	$(DOCKER) run --rm -v $(PWD)/.spectral.yaml:/tmp/.spectral.yaml:z -v $(PWD)/swagger.yaml:/tmp/swagger.yaml:z stoplight/spectral lint --ruleset "/tmp/.spectral.yaml" /tmp/swagger.yaml
+	$(DOCKER) run --rm -v $(PWD)/.spectral.yaml:/tmp/.spectral.yaml:z -v $(PWD)/swagger-backend.yaml:/tmp/swagger-backend.yaml:z stoplight/spectral lint --ruleset "/tmp/.spectral.yaml" /tmp/swagger.yaml
```
masayag, line 278: iinw, the path (the last argument) should be changed to `/tmp/swagger-backend.yaml` which is the path to which the yaml file is mounted in the image?

another suggestion is to run validation for both files within the same container. spectral usage should support several files as input.
```
spectral <command>

Commands:
  spectral lint [documents..]  lint JSON/YAML documents from files or URLs
```

swagger-backend.yaml
```diff
@@ -0,0 +1,184 @@
...
+    required:
+      - message
+    properties:
+      message:
+        type: string
+  device-configuration-response:
+    type: object
+    properties:
+      message:
+        type: string
```
masayag, line 26: pls add description for the properties.
it will be easier to understand what each `*-response` returns

masayag (changes requested):

swagger-backend.yaml
jordigilh: Understood

jordigilh: Done.

Makefile
jordigilh: Done.

swagger-backend.yaml
masayag: Thank you. this makes the API very clear :+1:

jakub-dzon (approved):

PR #242, opened 2022-06-01
add host devices types to hardware

tupyy:
This PR adds types for host devices. 

Linked to: https://issues.redhat.com/browse/ECOPROJECT-832

Signed-off-by: Cosmin Tupangiu <cosmin@redhat.com>


jakub-dzon (approved):
:+1:

machacekondra (approved):

api/v1alpha1/edgedevice_types.go
```diff
@@ -378,6 +381,27 @@ type SystemVendor struct {
...
+type HostDevice struct {
+
+	// path
+	Path string `json:"path,omitempty"`
+
+	// Device type
+	DeviceType string `json:"deviceType,omitempty"`
+
+	// owner id
+	UID int32 `json:"owner,omitempty"`
```
masayag, line 393: i can just assume that uint32 will be a better fit for this one and GID and version parts.
however, I don't recall why haven't we used it in other places when it made sense.

config/crd/bases/management.project-flotta.io_edgedevices.yaml
```diff
@@ -297,6 +297,34 @@ spec:
                           type: string
                       type: object
                     type: array
+                  hostDevices:
+                    description: host devices
```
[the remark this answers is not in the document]
masayag: +1, not just for this one, but also for the other properties below.
the description however will have to be added in [api/v1alpha1/edgedevice_types.go](https://github.com/project-flotta/flotta-operator/pull/242/files#diff-d751bc14a529b6b257da989fff2cafcd3797143688613e6fba8b8616ba27e46d)

api/v1alpha1/edgedevice_types.go
tupyy: my guess is to avoid mapping each value from int32 to int64. swagger generates only int64. 
I will change it since I'm making already a conversion from int32 -> int64

tupyy:
omg..

masayag (approved):

PR #244, opened 2022-06-01
Remove usage of finalizers on the EdgeDevice

jakub-dzon:
This PR removes the need for `EdgeDevice` finalizers - device deregistration is signalled by `EdgeDevice` being not present in 
the cluster or having `DeletionTimestamp` set.

Follow-up for https://github.com/project-flotta/flotta-operator/pull/236#discussion_r885375844

### DIFF
--- a/internal/edgeapi/backend/k8s/backend.go
+++ b/internal/edgeapi/backend/k8s/backend.go
@@ -42,6 +42,9 @@ func NewBackend(repository RepositoryFacade, assembler *ConfigurationAssembler,
 func (b *backend) GetRegistrationStatus(ctx context.Context, name, namespace string) (backendapi.RegistrationStatus, error) {
 	edgeDevice, err := b.repository.GetEdgeDevice(ctx, name, namespace)
 	if err != nil {
+		if errors.IsNotFound(err) {
+			return backendapi.Unregistered, nil
+		}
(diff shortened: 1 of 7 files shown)


internal/edgeapi/backend/k8s/backend.go
```diff
@@ -46,20 +42,17 @@ func NewBackend(repository RepositoryFacade, assembler *ConfigurationAssembler,
 func (b *backend) GetRegistrationStatus(ctx context.Context, name, namespace string) (backendapi.RegistrationStatus, error) {
 	edgeDevice, err := b.repository.GetEdgeDevice(ctx, name, namespace)
 	if err != nil {
+		if errors.IsNotFound(err) {
+			return backendapi.Unregistered, nil
```
masayag, line 46: Could it be that the device missing, but the EDSR was created?
So the actual status is pending approval?

jakub-dzon: It is possible that there is EDSR and not EdgeDevice, but `GetRegistrationStatus` would never be called in that situation - `control` requests are handled only after device is successfully registered and uses proper, device-specific, non-registration certificate.

masayag (approved):

tupyy (approved):

PR #245, opened 2022-06-01
Makefile: add make run-http-api

eloycoto:
This change adds a new Makefile target to run the http-api in local
without running the container, the main use cases:

- Define a Kubeconfig to be able to retrieve the right config on the API.
- Set the metrics port in different port, to be able to run with the
  controller.

I know that this makefile target is super opinionated, so please feel
free to close this PR if adds 0 value to you.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,6 +161,9 @@ run: manifests pre-build ## Run a controller from your host.
 	$(Q) kubectl create ns $(FLOTTA_OPERATOR_NAMESPACE) 2> /dev/null || exit 0
 	OBC_AUTO_CREATE=false ENABLE_WEBHOOKS=false LOG_LEVEL=debug go run -mod=vendor ./main.go
 
+http-api-run: ## Run HTTP API in localhost
+	METRICS_ADDR=":8089" go run cmd/httpapi/main.go
+
 docker-build: ## Build docker image with the manager.
 	$(DOCKER) build -t ${IMG} .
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -161,6 +161,10 @@ run: manifests pre-build ## Run a controller from your host.
 	$(Q) kubectl create ns $(FLOTTA_OPERATOR_NAMESPACE) 2> /dev/null || exit 0
 	OBC_AUTO_CREATE=false ENABLE_WEBHOOKS=false LOG_LEVEL=debug go run -mod=vendor ./main.go
 
+http-api-run: ## Run HTTP API in localhost
+	$(Q) export KUBECONFIG=~/.kube/config
```
masayag, line 165: in all other targets we don't require to define `KUBECONFIG`.
this seems to point on a missing functionality in `httpapi/main.go` to be able to look for config also on filesystem.
if you look at `make run` that invokes `go run ./main.go` it isn't needed.
[httpapi/main.go#getRestConfig](https://github.com/project-flotta/flotta-operator/blob/3c8fb521da78767750da972b7565be8b62dcf247/cmd/httpapi/main.go#L221) needs to be aligned with https://github.com/project-flotta/flotta-operator/blob/3c8fb521da78767750da972b7565be8b62dcf247/vendor/sigs.k8s.io/controller-runtime/pkg/client/config/config.go#L59

if you agree, I'd prefer to fix httpapi/main.go first.

pkliczewski (approved):

jakub-dzon (approved):

PR #247, opened 2022-06-06
Re-use existing model

jakub-dzon:
To avoid unnecessary manual type conversions/mapping I've used existing model generated from swagger for backend communication.


masayag (approved):

swagger-backend.yaml
```diff
@@ -197,6 +209,44 @@ paths:
           description: Error
           schema:
             $ref: '#/definitions/error'
+  '/namespaces/{namespace}/devices/{device-id}/registration/namespace':
```
jordigilh, line 212: I'm not entirely sure this is the right approach. Even if you wanted to go forth with it, I think it would make more sense for this to be:
`.../registration?namespace`
since you're retrieving a field/attribute of the object.
Maybe return the namespace as part of the GET or PUT return model operation? You'll save an extra call in this case :smile:

For GET:

```yaml
  device-registration-status-response:
    type: object
    properties:
      status:
        type: string
        description: >-
          Returns the device registration status, which can be one of the
          following {registered, unregistered, unknown}.
      namespace:
        type: string
        description: >-
          Returns the target namespace where the device can be found.
      message:
        type: string
        description: >-
          Exposes the error message generated at the backend when there is an
          error (example HTTP code 500).
```

swagger-backend.yaml
```diff
@@ -41,7 +41,11 @@ definitions:
           Exposes the error message generated at the backend when there is an
           error (example HTTP code 500).
       device-configuration:
-        $ref: 'swagger.yaml#/definitions/device-configuration-message'
+        x-go-type:
```
jordigilh, line 44: This is great! :smile:

swagger-backend.yaml
jakub-dzon: Hmmm, that's an interesting approach with using GetRegistrationStatus. Thanks!

jordigilh (approved):

masayag (approved):

PR #251, opened 2022-06-07
Option to use remote backend

jakub-dzon:
This PR adds support for choosing remote HTTP backend in the Edge API server instead of default k8s-based one.

To use remote backend, following env property needs to be set for the Edge API server: `REMOTE_BACKEND_URL`. If URL uses `https` schema, Edge API Server's TLS certificates will be used (same as ones used for handling agent connections).

### DIFF
--- a/internal/edgeapi/backend/factory/factory.go
+++ b/internal/edgeapi/backend/factory/factory.go
@@ -2,6 +2,7 @@ package factory
 
 import (
 	"crypto/tls"
+	"fmt"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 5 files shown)


internal/edgeapi/backend/factory/factory.go
```diff
@@ -1,44 +1,88 @@
...
+type Factory struct {
+	InitialDeviceNamespace string
+	Logger                 *zap.SugaredLogger
+	Client                 kubeclient.Client
+	EventRecorder          record.EventRecorder
+	TLSConfig              *tls.Config
+}
+
+func (f *Factory) Create(remoteBackendURL string, remoteBackendTimeout time.Duration) (backend.EdgeDeviceBackend, error) {
+	if remoteBackendURL == "" {
```
gabriel-farache, line 37: This is too specific IMO, if we have to add new backend, we cannot only check some env; what will happen if some backend share the same parameters?
I would do something like that:
```

backendsMap := map[string]interface{}{
        "k8s":   f.createK8sBackend,
        "remote": f.createRemoteBackend,
    }
func ParseBackend(str string) (interface{}, bool) {
    c, ok := backendsMap[strings.ToLower(str)]
    return c, ok
}
```
(inspiration taken from [here](https://stackoverflow.com/questions/68543604/best-way-to-parse-a-string-to-an-enum))
And then you just call the `ParseBackend` with the (new) input argument explicitly selecting the backend.
I take as assumption that the method creating the backend will have a way of accessing the env var/arguments given to the program (so maybe we would need to add a config object that gather and give access to the env var needed to run)
WDYT?

jakub-dzon: Hmm, being explicit about the backend choice has its appeal and I'll have a look how that would look like. I didn't feel strong about neither option.

jakub-dzon: Please have a look now.

gabriel-farache (approved):
\lgtm
\cat

PR #252, opened 2022-06-07
Added EdgeDevice finalizer to ensure EDSR removal

jakub-dzon:
This PR introduces a new `EdgeDevice` finalier to ensure that related `EDSR` is removed. 
This finalizer _does not_ directly block device deregistration.

Signed-off-by: Jakub Dzon <jdzon@redhat.com>

PR #253, opened 2022-06-08
Replacing reconcile-based EDSR clean-up with owner reference

jakub-dzon:
This PR reverts #252 and replaces reconcile-based `EDSR` clean-up on `EdgeDevice` removal.


jakub-dzon:
This approach is not feasible in our case - cross-namespace owner references are not allowed and `EdgeDevice` and `EDSR` may (and probably often will) end up in different namespaces.

PR #259, opened 2022-06-10
Rename system-metrics-configuration for generic reuse

jordigilh:
Renamed to component-metrics-configuration to be able to share it with datatransfer-metrics-configuration, responsible for retrieving the data metrics generated in the agent for data synchronization.

This PR depends on https://github.com/seqsense/s3sync/pull/255


jakub-dzon (commented):
It looks good; please make the GH actions happy.

.github/workflows/go.yml
```diff
@@ -78,4 +78,4 @@ jobs:
       - name: Run Gosec Security Scanner
         uses: securego/gosec@master
         with:
-          args: '-exclude-generated  ./...'
```
jakub-dzon, line 81: Wouldn't `-no-fail` ignore scan failures? I don't think we want that.

jakub-dzon: I see in the logs of that action that it failed, but the action is green:
```
 Golang errors in file: []:
  > [line 0 : column 0] - error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```

jordigilh: Yep, it won't fail but it will scan. I posted in slack my take on this. Waiting for feedback on how we should address this. Just wanted to prove this was an option.

jakub-dzon (approved):

pkliczewski (approved):

PR #262, opened 2022-06-15
Fix wrong workload status on device after deletion

gabriel-farache:
Fix the wrong status  of a workload after its deletion: the workload should not be in edgedevice resource after it has been deleted

Current:
```
name: workload
phase: Deploying
```

After fix:
`null`

Signed-off-by: Gabriel Farache gfarache@redhat.com

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -104,6 +104,26 @@ var _ = Describe("e2e", func() {
 			Expect(err).To(BeNil())
 			Expect(stdout).To(ContainSubstring("Welcome to nginx!"))
 		})
+		It("Deploy then remove valid edgeworkload to registered device", func() {
+			// given
+			err := device.Register()
+			Expect(err).To(BeNil())
+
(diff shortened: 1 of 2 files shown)


jakub-dzon (approved):

test/e2e/e2e_test.go
```diff
@@ -104,6 +104,26 @@ var _ = Describe("e2e", func() {
 			Expect(err).To(BeNil())
 			Expect(stdout).To(ContainSubstring("Welcome to nginx!"))
 		})
+		It("Deploy then remove valid edgeworkload to registered device", func() {
+			// given
+			err := device.Register("dnf install ansible -y")
```
machacekondra, line 109: I think you can remove here this `dnf` command

machacekondra (approved):

test/e2e/e2e_test.go
gabriel-farache: done

PR #269, opened 2022-06-22
Migrate to go 1.18

jordigilh:
* Bumped go version 1.18
* Bumped mockery to 2.13.1 (v2):
  * Mockery version 1.x is not compatible with golang 1.18. Version 2 however include changes to the CLI that are not backwards compatible, thus causing the generated files from go-swagger to fail to generate when invoking go generate ./... The commit includes the go-swagger templates for the client and server that containt the correct mockery command flags.
  * Bumped references to golang 1.18 in the Makefile to handle the import target
* Regenerated files from running go-swagger and mockery


jordigilh:
> At some point, I think that we discussed this in the past, and it was an issue with the code imported by device worker. , so I highly suggest that check that import is working as expected.

Tested device worker against 1.18 without changes to the flotta-operator package version and it seems to work fine. I did run `go get -u ./.../` but it only changed the indirect dependencies. Do you remember where did you find the issues?
[Here](https://github.com/project-flotta/flotta-device-worker/pull/180) is the PR for device-worker with go 1.18.

jordigilh:
Summary of this PR:
* The main bulk is changes in the go.mod and dependencies to migrate to go 1.18, including bumping mockery version to the latest (currently using 1.1.2, which is deprecated). Many files changed, blame `go mod tidy` and `go mod download` :smile:, I'm just the messenger :wink: 
* Fixes the gosec issues with go 1.18 that we faced earlier and we patched by adding the flag to not to fail when error. However I discovered that to fix this issue we don't require to migrate the project to go 1.18, just add the env variable `GOFLAGS: -buildvcs=false` in the step and it take care of it.

Are we comfortable migrating to go 1.18 and including the fix for gosec? or should we take them apart and merge them independently?

And I tested migrating the device-worker to go 1.18 and it was quite a straight task: https://github.com/project-flotta/flotta-device-worker/pull/180

Thoughts? @eloycoto @jakub-dzon @machacekondra @pkliczewski @gabriel-farache @tupyy @masayag

masayag:
> > At some point, I think that we discussed this in the past, and it was an issue with the code imported by device worker. , so I highly suggest that check that import is working as expected.
> 
> Tested device worker against 1.18 without changes to the flotta-operator package version and it seems to work fine. I did run `go get -u ./.../` but it only changed the indirect dependencies. Do you remember where did you find the issues? [Here](https://github.com/project-flotta/flotta-device-worker/pull/180) is the PR for device-worker with go 1.18.

IIRC, @tupyy encountered compatibility issue with device-worker and go-1.18

masayag:
> Summary of this PR:
> 
> * The main bulk is changes in the go.mod and dependencies to migrate to go 1.18, including bumping mockery version to the latest (currently using 1.1.2, which is deprecated). Many files changed, blame `go mod tidy` and `go mod download` smile, I'm just the messenger wink
> * Fixes the gosec issues with go 1.18 that we faced earlier and we patched by adding the flag to not to fail when error. However I discovered that to fix this issue we don't require to migrate the project to go 1.18, just add the env variable `GOFLAGS: -buildvcs=false` in the step and it take care of it.
> 
> Are we comfortable migrating to go 1.18 and including the fix for gosec? or should we take them apart and merge them independently?
> 
> And I tested migrating the device-worker to go 1.18 and it was quite a straight task: [project-flotta/flotta-device-worker#180](https://github.com/project-flotta/flotta-device-worker/pull/180)
> 
> Thoughts? @eloycoto @jakub-dzon @machacekondra @pkliczewski @gabriel-farache @tupyy @masayag

I'd break into two pieces (one to handle gosec which can be merged instantly, and a second PR for 1.18)

tupyy:
> > > At some point, I think that we discussed this in the past, and it was an issue with the code imported by device worker. , so I highly suggest that check that import is working as expected.
> > 
> > 
> > Tested device worker against 1.18 without changes to the flotta-operator package version and it seems to work fine. I did run `go get -u ./.../` but it only changed the indirect dependencies. Do you remember where did you find the issues? [Here](https://github.com/project-flotta/flotta-device-worker/pull/180) is the PR for device-worker with go 1.18.
> 
> IIRC, @tupyy encountered compatibility issue with device-worker and go-1.18

It is not `go1.18` per se the problem. It is the standard library that can cause some issues. There are small differences between std lib 1.17 and 1.18 and we need to be careful using dependencies that don't support 1.18

gabriel-farache:
> > Summary of this PR:
> > 
> > * The main bulk is changes in the go.mod and dependencies to migrate to go 1.18, including bumping mockery version to the latest (currently using 1.1.2, which is deprecated). Many files changed, blame `go mod tidy` and `go mod download` smile, I'm just the messenger wink
> > * Fixes the gosec issues with go 1.18 that we faced earlier and we patched by adding the flag to not to fail when error. However I discovered that to fix this issue we don't require to migrate the project to go 1.18, just add the env variable `GOFLAGS: -buildvcs=false` in the step and it take care of it.
> > 
> > Are we comfortable migrating to go 1.18 and including the fix for gosec? or should we take them apart and merge them independently?
> > And I tested migrating the device-worker to go 1.18 and it was quite a straight task: [project-flotta/flotta-device-worker#180](https://github.com/project-flotta/flotta-device-worker/pull/180)
> > Thoughts? @eloycoto @jakub-dzon @machacekondra @pkliczewski @gabriel-farache @tupyy @masayag
> 
> I'd break into two pieces (one to handle gosec which can be merged instantly, and a second PR for 1.18)

I second that

About the regression we may face with the device-worker, I guess you ran the e2e tests so if there was no error we should be fine. And if we are not, this will be the occasion to improve the e2e tests are they are supposed give us confidence on impact of changes on the operator to the overall workflow

Overall, it's a good idea to keep up with the new version once it's been released for a few month (to see if there is anyu bugs), even more with go 1.18 that brings generics!

eloycoto:
TBH, with the low resources that the project will have over the next months and the possible problems that 1.18 dependencies will bring to the device worker until I saw a big change that will make our life much better, I'll keep Golang version as it is. IMHO.

eloycoto:
> Overall, it's a good idea to keep up with the new version once it's been released for a few month (to see if there is any bugs), even more with go 1.18 that brings generics!

Not sure if generics, as it is now, is useful for flotta at all. There are use cases sure, but without generics on structs methods I do not think that adds too much value as the project is now. IMHO

PR #271, opened 2022-06-23
Install: FDO integration 🛬

eloycoto:
This adds a better way to handle rootless podmand on rpm-ostree based
system and how to use FDO GUID instead of the machine-id.


hack/install-agent-rpm-ostree.sh.template
```diff
@@ -61,6 +60,15 @@ cat <<EOF > /etc/pki/consumer/cert.pem
 <CERT_PEM>
 EOF
 
+
+function get_client_id() {
+  if [ -f "/etc/device-credentials" ]; then
+    /usr/libexec/fdo/fdo-owner-tool dump-device-credential /etc/device-credentials | grep "Device GUID" | awk '{print $NF}'
```
jakub-dzon, line 66: Is that command available by default in RHEL4Edge?

eloycoto: yes, if the customizatoin-fdo is in there, it'll be in there.  this is why I checked the file.

jakub-dzon: The other template doesn't have that `if` - does it meant that `fdo-owner-tool` is not available in other, dnf-supporting OSes?

eloycoto: fdo-owner-tool is only availbe on rhel edge right now, fedora iot not yet.

jakub-dzon (approved):

hack/install-agent-rpm-ostree.sh.template
```diff
@@ -93,8 +102,8 @@ chmod 644 /etc/systemd/system/init-yggd.service
 
 systemctl enable init-yggd.service
 
-ln -s /usr/lib/systemd/system/podman.service /etc/systemd/system/default.target.wants/podman.service
-ln -s /usr/lib/systemd/system/podman.socket /etc/systemd/system/sockets.target.wants/podman.socket
+systemctl enable podman.service
+systemctl enable podman.socket
```
masayag, line 106: question: why change only for podman and not for the other services below?

eloycoto: Podman is already in there, part of the rhel-edge, but others ones cannot be enabled because does not exist yet!

masayag (approved):

pkliczewski (approved):

PR #272, opened 2022-06-24
Added knobs to enable running workloads by rooted podman

jordigilh:
* Modified the EdgeWorkload CRD to include a new subsection named `SecurityContext`. This section contains a new boolean knob named `RunByRoot` that, when set to true, will instruct the agent to run this workload with podman's root context. By default, when false or not specified, the workload runs in the flotta's podman user space.
* Added a new knob in the swagger model for the workloads to expose the `RunByRoot` knob previously mentioned
* Modified edgedevice reconciliation process to include the security context when generating the `models.Workload` structure
* Regenerated files from running `make generate`
* Added unit tests


api/v1alpha1/edgeworkload_types.go
```diff
@@ -33,13 +33,20 @@ type EdgeWorkloadSpec struct {
...
 	// from this endpoint. This key is what is defined on the edgedevice
 	// logCollection property
 	LogCollection string `json:"logCollection,omitempty"`
 }
 
+// EdgeWorkloadSecurityContext holds workload-level security attributes
+type EdgeWorkloadSecurityContext struct {
+	// RunByRoot determines if the containers in the workload should run under the root user. By default
+	// workloads are run by a non-root user.
+	RunByRoot *bool `json:"runByRoot,omitempty"`
```
masayag, line 48: Is the demand only to be run as root or as flotta user or will there be a case for "RunAs" and provide the user that should execute this workload?

config/crd/bases/management.project-flotta.io_edgeworkloads.yaml
```diff
@@ -7207,6 +7207,16 @@ spec:
                 required:
                 - spec
                 type: object
+              securityContext:
+                description: EdgeWorkloadSecurityContext holds workload-level security
+                  attributes
+                properties:
+                  runByRoot:
```
masayag, line 7214: At the [API](https://github.com/project-flotta/flotta-operator/pull/272/files#diff-8b1949772e223a1da6a2049ada2733fa506410975b241cf86cf44c7a8665bc62R782) level this field is called `RunAsRoot`. Why the difference? (I like `RunAsRoot` since it doesn't reveal information of how this will be executed)

api/v1alpha1/edgeworkload_types.go
jordigilh: As far as I know either as root or as the flotta user.

config/crd/bases/management.project-flotta.io_edgeworkloads.yaml
tupyy: Personally, I would use the same convention as `podman` `rootful`. my 2 cents

eloycoto:
Then I do not fully understand this PR, if we use it like that, why we will use rooted podman instead of --privilege and secoomp profiles?

I mean, I cannot see the difference at all, and the podspec is what people use it.

jordigilh:
> Then I do not fully understand this PR, if we use it like that, why we will use rooted podman instead of --privilege and secoomp profiles?
> 
> I mean, I cannot see the difference at all, and the podspec is what people use it.

My understanding is that adding `--privilege` for a podman running as rootless will not grant the container any extra privilege that is not available to the podman's instance already. I presume that seccomp will be equaly restricted based on the seccomp flags enabled in the podman's instance.

My understanding of the motivations for having a rootful podman is to allow containers to access specific hardware that can only be done via a rooted container runtime.

But if we can achieve this goal via a rootless podman, honestly, it's probably better to take a look at it, because right now it's looking a bit messy on the agent side: Because we have 2 podman's running, we have to make sure that a given workload can only be run once, even though it could run in 2 podman's at the same time, or that the generated pod ID does not have a collision in the internal name-podID mapping).

config/crd/bases/management.project-flotta.io_edgeworkloads.yaml
jordigilh: By doing that you're exposing implementation details and needs to be populated on each workload. Having a boolean with false by default simplifies the workload spec since the lack of it in the manifest will always mean that it is run by the flotta user, which is by probably the more generic use case.

jordigilh:
> Quick one, why using this instead of the privilege and secomp profiles in the PodSpec?

The issue here is that internally we target 2 different podman instances. If we were to target a single one, the solution you propose to use the pod spec's security context would be sufficient.

eloycoto:
If the problem is the hardware, then use the udev rules, no?

jordigilh:
> If the problem is the hardware, then use the udev rules, no?

in that case, it might work. It would require to create the rules as part of the deployment, maybe using ansible?, but it would do the trick. What do you think @pkliczewski @masayag @tupyy ?

pkliczewski:
> in that case, it might work. It would require to create the rules as part of the deployment, maybe using ansible?, but it would do the trick. What do you think @pkliczewski @masayag @tupyy ?

Do you want to do this change in this PR or create a new one?

tupyy:
> > in that case, it might work. It would require to create the rules as part of the deployment, maybe using ansible?, but it would do the trick. What do you think @pkliczewski @masayag @tupyy ?
> 
> Do you want to do this change in this PR or create a new one?

If the tests with `udev rules` are ok (I don't see why not), there will be no change in CRD. The user will use his own playbook to modify the rules.
IMO, we should keep this PR open until we validate @eloycoto proposal.

PR #273, opened 2022-06-29
CI: Fix e2e flakes

eloycoto:
With the new changes on RPM/systemd/sysusers the e2e was failing hard.
This commit adds the following changes:

- Avoid the use of machine-ctl and use sudo -u flotta
- Forcing endpoint container to have CgroupNsMode to private.
- Separate kind phases to make it easier to debug failures.
- Gatter better logs

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -70,7 +70,7 @@ var _ = Describe("e2e", func() {
 			Expect(stdout).To(Equal("active"))
 
 			// Check the podman.socket is running:
-			stdout, err = device.Exec("systemctl --machine flotta@.host is-active --user podman.socket")
+			stdout, err = device.Exec("sudo -u flotta systemctl is-active --user podman.socket")
 			Expect(err).To(BeNil())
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_test.go
```diff
@@ -70,7 +70,7 @@ var _ = Describe("e2e", func() {
 			Expect(stdout).To(Equal("active"))
 
 			// Check the podman.socket is running:
-			stdout, err = device.Exec("systemctl --machine flotta@.host is-active --user podman.socket")
```
jordigilh, line 73: I'm curious, why the change?

jordigilh (approved):

test/e2e/e2e_test.go
eloycoto: mainly, because all systemd env vars that are needed, so checking as flotta user using sudo it's much simpler.

machacekondra: Strange that the command don't fail on missing env vars. I thought sudo pam modules don't create the needed env vars.

PR #285, opened 2022-07-14
Detect version from os-release

masayag:


hack/install-agent-dnf.sh.template
```diff
@@ -17,19 +17,17 @@ OPTIONS:
            to make the Flotta API accessible from your machine
    -p      Port to connect to Flotta API, default 8043
    -t      Indicates if to use testing repository, default is false
-   -v      Specifies the Fedora version to install, default is 35
```
gciavarrini, line 20: Do we need to document the default value somewhere?

masayag: no, it isn't needed. the version will be detected for each fedora. the user of the script shouldn't care about it.

PR #300, opened 2022-09-19
Block workloads with `HostPort` 9100

### LINKED ISSUE #281: Block workloads with hostPort 9100
Port 9100 on devices is reserved for node_exporter service:
```
[root@fedora ~]# netstat -tunlp | grep node_exporter
tcp        0      0 127.0.0.1:9100          0.0.0.0:*               LISTEN      917/node_exporter
```
Therefore validation is required for edge-workload not to create one using this port.

arielireni:
This PR fixes #281.

### DIFF
--- a/api/v1alpha1/edgeworkload_webhook.go
+++ b/api/v1alpha1/edgeworkload_webhook.go
@@ -101,6 +101,14 @@ func (r *EdgeWorkload) validate() error {
 		} else {
 			containersNames[container.Name] = struct{}{}
 		}
+
+		if container.Ports != nil {
+			for _, port := range container.Ports {
+				if port.HostPort == 9100 {
+					return fmt.Errorf("HostPort 9100 is reserved for internal use on the device and cannot be set for user workloads")
+				}
+			}
+		}
 	}
 
 	for _, volume := range podSpec.Volumes {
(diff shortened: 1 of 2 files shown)


api/v1alpha1/edgeworkload_webhook.go
```diff
@@ -101,6 +101,12 @@ func (r *EdgeWorkload) validate() error {
 		} else {
 			containersNames[container.Name] = struct{}{}
 		}
+
+		for _, port := range container.Ports {
+			if port.HostPort == 9100 {
+				notValidPaths = append(notValidPaths, containersMsg(container, "ports"))
```
masayag, line 107: Which message will be presented in a case of failure?

masayag:
Pls have a look at the failed CI issue

api/v1alpha1/edgeworkload_webhook.go
gciavarrini, line 107: nit: to be consistent with the other if blocks, I think `ports` may be replaced with `port`

arielireni: I agree with the need to be consistent, but WDYT about `ports.hostPort`? to be consistent with lines 66-85

masayag:
@arielireni pls rebase

api/v1alpha1/edgeworkload_webhook.go
arielireni: `the following paths in podSpec are not supported and should be removed: containers[container.Name].hostPort`

masayag: So this error message isn't accurate. It should say something like: "HostPort 9100 is reserved for internal use on the device and cannot be set for user workloads"

arielireni: done

masayag (approved):

